PR #278, opened 2024-05-01
Add "glide tag list" support

praveingk:
Returns tag mappings 
Example : 

```
$ glide tag list
0). tag_name:"ibmvm1"  child_tags:"default.ibm.vm-1"
1). tag_name:"default.ibm.vm-1"  uri:"/resourcegroup/2d6e51c90e42424b8757d5a481e2450e/zone/us-east-1/instance/0757_7764d121-6465-4c70-a791-6fce774dd2fb"  ip:"10.0.0.9"
2). tag_name:"prod"  child_tags:"ibmvm1"
3). tag_name:"ibmvm2"  child_tags:"default.ibm.vm-2"
4). tag_name:"default.ibm.vm-2"  uri:"/resourcegroup/2d6e51c90e42424b8757d5a481e2450e/zone/us-east-1/instance/0757_006cd7db-1795-4b68-a380-1a6f023b8f65"  ip:"10.0.0.10"
```

### DIFF
--- a/pkg/tag_service/tag_service.go
+++ b/pkg/tag_service/tag_service.go
@@ -251,7 +251,7 @@ func (s *tagServiceServer) ListTags(c context.Context, e *emptypb.Empty) (*tagse
 		tagMapping, err := s.GetTag(c, &tagservicepb.Tag{TagName: tag})
 		if err != nil {
 			// Ignore errors
-			utils.Log.Println("Failed to get tag mapping of %s: %v", tag, err)
+			utils.Log.Printf("Failed to get tag mapping of %s: %v\n", tag, err)
 			continue
 		}
 		resolvedTagList = append(resolvedTagList, tagMapping)
--- a/pkg/tag_service/tag_service.go
+++ b/pkg/tag_service/tag_service.go
@@ -29,6 +29,7 @@ import (
 	redis "github.com/redis/go-redis/v9"
 
 	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/emptypb"
 )
 
 type tagServiceServer struct {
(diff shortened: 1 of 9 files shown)


divega (approved):
LGTM, but don't rely just on my review.

pkg/tag_service/tag_service.go
```diff
@@ -241,6 +242,22 @@ func (s *tagServiceServer) ResolveTag(c context.Context, tag *tagservicepb.Tag)
 	return &tagservicepb.TagMappingList{Mappings: resolvedTags}, nil
 }
 
+// Resolve a list of tags into all base-level IPs
+func (s *tagServiceServer) ListTags(c context.Context, e *emptypb.Empty) (*tagservicepb.TagMappingList, error) {
+	var resolvedTagList []*tagservicepb.TagMapping
+	tags := s.client.Keys(c, "*").Val()
+	for _, tag := range tags {
+		tagMapping, err := s.GetTag(c, &tagservicepb.Tag{TagName: tag})
+		if err != nil {
```
smcclure20, line 251: I'm fine with this for now, but should we give a warning if there's an error? Or, maybe we should error out of all requests give an error (since that may mean the tag service is down or something)

smcclure20 (commented):
Looks good, just one little note on errors. Thanks for adding this!

pkg/tag_service/tag_service.go
praveingk: I have added a log, since returning error is not ideal for IBM plugin, where KV store use the same DB, and it would fail with logs like this : .
`GetTag default:ibm:r014-af7745bc-38d7-4f9e-9957-07c022ddcf7a: GetTag default:ibm:r014-af7745bc-38d7-4f9e-9957-07c022ddcf7a: WRONGTYPE Operation against a key holding the wrong kind of value"}null`

Hence, I have just added a log, and continuing to resolve other tags.

PR #284, opened 2024-05-02
[Infra] Add git hook for sign-off

seankimkdy:
Added a git hook so contributors don't have to `git commit -s` every time.

### DIFF
--- a/.githooks/prepare-commit-msg
+++ b/.githooks/prepare-commit-msg
@@ -1,12 +1,5 @@
 #!/bin/sh
 
-# Check for an empty commit message
-if [ -z "$2" ]
-then
-    echo "Empty commit message is not allowed."
-    exit 1
-fi
-
 # Add sign-off
 NAME="$(git config user.name)"
 EMAIL="$(git config user.email)"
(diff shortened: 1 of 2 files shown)


.githooks/prepare-commit-msg
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+# Check for an empty commit message
+if [ -z "$2" ]
+then
+    echo "Empty commit message is not allowed."
```
praveingk, line 6: I tried this, 
seems like it would enforce usage of "git commit -m <files>"
rather than "git commit <files>"  which traditionally would spawn an editor to add the commit message.  
Just a matter of preference, but maybe worth mentioning it in after the commands to install the hook.

praveingk (changes requested):

.githooks/prepare-commit-msg
seankimkdy: Good point. I removed it. It seems like git catches the empty message regardless, so this was useless to begin with.

smcclure20 (approved):

praveingk (approved):

PR #287, opened 2024-05-02
Create dependabot.yml

### LINKED ISSUE #285: Set up dependabot to keep dependencies up to date
Dependabot can automate the generation of PRs like #146:

https://docs.github.com/en/code-security/dependabot

One caveat for an apparently we are not supposed to use  go.sum: https://github.blog/changelog/2023-03-07-dependency-graph-removes-go-sum-support/.

smcclure20:
Resolves #285

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,23 @@
+# To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
+# Please see the documentation for all configuration options:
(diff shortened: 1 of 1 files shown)


seankimkdy (approved):

.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
+# To get started with Dependabot version updates, you'll need to specify which
```
divega, line 1: Nit: should we clean this up/remove comments?

divega (approved):

PR #288, opened 2024-05-03
Revert title of index page to Welcome to Paraglider and remove logo

divega:
Otherwise there are two logos side by side on index.


smcclure20 (commented):
I'm okay with this, though tbh I didn't think the double logo for the homepage was that bad. So I'll defer to others on this.

seankimkdy (approved):
Looks fine to me

PR #290, opened 2024-05-09
[Azure] Add option to clean up resources without deleting resource group

### LINKED ISSUE #289: [Azure] Alternative Resources Cleanup
Hey team,
I'm having a hard time debugging IBM-Azure VPN connectivity due to Azure's plugin cleanup policy. 
My team doesn't get privileges on the Azure subscription level, but rather each team gets ownership over an Azure resource group, hence creating/deleting resource groups isn't possible.
With that in mind, I can't terminate test resources efficiently, forced to delete resources one-by-one, in a specific order, waiting for child resources to delete before going forward with deleting their parent resource.

I'm not sure we can rely on future users/devs to:
- Have full privileges over the entire subscription of their Azure account (to allow resource group creation/deletion).
- Have a resource group for themselves. Test cleanup would delete other team member's resources.

I would appreciate if the Azure plugin could implement a cleanup function that strictly deletes Paraglider resources matching specified parameters, such as a namespace/region.

Thanks.

seankimkdy:
Resolves #289.

I added an environment variable `INVISINETS_AZURE_RESOURCE_GROUP` where you can specify the resource group you'd like used. I didn't make it with `PARAGLIDER` prefix because all other environment variables are currently `INVISINETS` prefix. I'll put up another PR to rename everything.

I introduced a deletion order (see `utils.go`) based on an educated guess. It seems to work fine so far, but may have to be adjusted if we add more resource types.

The deletion is by filtering for resources within a given resource group and checking an Azure tag called `paraglider-deployment`. I edited the `sdk_handler.go` to add this tag wherever applicable (for most high-level resources).

Note that this would be used for all tests meaning the 2 integration tests (within Azure) and 1 multicloud test would all run within the resource group. This is fine for now since these test run sequentially but is something we'll need to consider if we ever run these tests concurrently. Theoretically, we can run all the tests concurrently if we make the namespaces and VM names unique.

Our CI pipeline will stay on the existing paradigm of creating/deleting resource groups per test.


pkg/azure/plugin_integration_test.go
```diff
@@ -90,44 +91,46 @@ func TestBasicPermitListOps(t *testing.T) {
...
 
 	// Delete permit list rule
-	deletePermitListResp, err := s.DeletePermitListRules(ctx, &paragliderpb.DeletePermitListRulesRequest{RuleNames: []string{rules[0].Name}, Namespace: "default", Resource: vmID})
+	deletePermitListResp, err := s.DeletePermitListRules(ctx, &paragliderpb.DeletePermitListRulesRequest{RuleNames: []string{rules[0].Name}, Namespace: namespace, Resource: vmID})
 	require.NoError(t, err)
 	require.NotNil(t, deletePermitListResp)
 
 	// Assert the rule is deleted by using the get permit list api
-	getPermitListResp, err = s.GetPermitList(ctx, &paragliderpb.GetPermitListRequest{Resource: vmID, Namespace: "default"})
+	getPermitListResp, err = s.GetPermitList(ctx, &paragliderpb.GetPermitListRequest{Resource: vmID, Namespace: namespace})
```
divega, line 111: Curious if changes like this are also a bug fix.

pkg/azure/resources.go
```diff
@@ -33,8 +33,17 @@ import (
 )
 
 const (
-	virtualMachineTypeName = "Microsoft.Compute/virtualMachines"
-	managedClusterTypeName = "Microsoft.ContainerService/managedClusters"
+	virtualMachineTypeName        = "Microsoft.Compute/virtualMachines"
```
divega, line 36: Curious if these constants was supposed to be only for supported compute resource types or if it makes sense to extend it to all the resource types that the plugin manages. I understand that these aren't exported so they are just an implementation detail, but I take all the existing plugins a sample code for people that want to implement more plugins, so I am curious if for these we are following similar pattern across plugins.

divega (approved):
LGTM, but I don't have the full context.

pkg/azure/resources.go
seankimkdy: > these constants was supposed to be only for supported compute resource types

Yea I think that was the intent but I couldn't find a better home for the rest of them. Open to ideas or leaving it for the time being. This doesn't really exist in other plugins as far as i know.

pkg/azure/plugin_integration_test.go
seankimkdy: Nope I just made it into a variable because I made a function that takes in the namespace as an arg.

pkg/azure/resources.go
```diff
@@ -33,8 +33,17 @@ import (
...
+	managedClusterTypeName        = "Microsoft.ContainerService/managedClusters"
```
smcclure20, line 37: Hm, the previous ones were here since the rest of the code in this file needs to tell those two resources apart. These other resource types are not relevant to anything else in this file. Could this all go in a utils thing instead?

pkg/azure/sdk_handler.go
```diff
@@ -548,6 +551,7 @@ func (h *AzureSDKHandler) CreateParagliderVirtualNetwork(ctx context.Context, lo
 			},
 		},
 	}
+	h.createParagliderNamespaceTag(&parameters.Tags)
```
smcclure20, line 554: I am not opposed to this tagging way of doing things, but before our standard has been to determine the namespace of a resource based on what vnet it is in / a naming convention. Is there a reason that doesn't work here?

pkg/azure/utils.go
```diff
@@ -67,20 +75,133 @@ func SetupAzureTesting(subscriptionId string, testName string) string {
...
+									resourceTypeToAPIVersion[providerNamespace+"/"+*resourceType.ResourceType] = *resourceType.APIVersions[0] // Use most recent API version
+								}
+							}
+						}
+						resourceTypeToResources[*resource.Type] = append(resourceTypeToResources[*resource.Type], resource)
+					}
+
+				}
+				// Delete resources in the following order
+				// TODO @seankimkdy: figure out where clusters fit in on this order
```
smcclure20, line 156: They should be at the same step as VMs

pkg/azure/utils.go
```diff
@@ -67,20 +75,133 @@ func SetupAzureTesting(subscriptionId string, testName string) string {
...
+					resourceIDInfo, err := getResourceIDInfo(*resource.ID)
+					if err != nil {
+						panic(fmt.Sprintf("Unable to parse resource ID info: %v", err))
+					}
+
+					// Need to check here since resourceGroup can't be used with tags in the above filter
+					// Ignore case when checking because Azure sometimes capitalizes the resource group name for no apparent reason
+
+					if strings.EqualFold(resourceIDInfo.ResourceGroupName, resourceGroupName) {
+						if _, ok := resourceTypeToResources[*resource.Type]; !ok {
```
smcclure20, line 130: Nit: Would have helped me reading this to have a comment like "Group resources by type and get the correct API version for the type" or something like that

pkg/azure/utils.go
```diff
@@ -67,20 +75,133 @@ func SetupAzureTesting(subscriptionId string, testName string) string {
...
+					localNetworkGatewayTypeName,
+					publicIPAddressTypeName,
+					virtualNetworkTypeName,
+					networkSecurityGroupTypeName,
+					networkWatcherTypeName,
+				}
+				for _, resourceType := range deletionOrder {
+					if resources, ok := resourceTypeToResources[resourceType]; ok {
+						err = deleteResources(ctx, resourcesClient, resources, resourceTypeToAPIVersion[resourceType])
+						if err != nil {
```
smcclure20, line 172: We'll have to see if this is necessary (especially since this is to help with manual testing anyways), but if the failure rate for deletions is high enough, it might be worth having a retry mechanism here

smcclure20 (commented):
Just a few comments, otherwise ready to approve

pkg/azure/sdk_handler.go
seankimkdy: It just seemed like more than it's worth to find all resources created for a test. It also wouldn't cover resources like the vnet that the VPN gateway resides in (at the very least there would be a lot of pointer chasing). I felt this was the most straightforward way to do this in the fastest time.

pkg/azure/resources.go
seankimkdy: Yea I'll throw them in `utils.go` instead

pkg/azure/utils.go
seankimkdy: Yea I'll keep that in the back of my mind. We can cross that bridge if more people use this option and it becomes an issue.

pkg/azure/resources.go
```diff
@@ -33,17 +33,8 @@ import (
...
-	localNetworkGatewayTypeName   = "Microsoft.Network/localNetworkGateways"
-	diskTypeName                  = "Microsoft.Compute/disks"
-	connectionTypeName            = "Microsoft.Network/connections"
-	networkInterfaceTypeName      = "Microsoft.Network/networkInterfaces"
-	networkSecurityGroupTypeName  = "Microsoft.Network/networkSecurityGroups"
-	publicIPAddressTypeName       = "Microsoft.Network/publicIPAddresses"
-	virtualNetworkGatewayTypeName = "Microsoft.Network/virtualNetworkGateways"
-	virtualNetworkTypeName        = "Microsoft.Network/virtualNetworks"
-	networkWatcherTypeName        = "Microsoft.Network/networkWatchers"
+	virtualMachineTypeName = "Microsoft.Compute/virtualMachines"
```
smcclure20, line 36: Not to require more iterations than needed, but I wouldn't be opposed to having these in the utils too.

smcclure20 (approved):

Cohen-J-Omer (approved):
Tested it on basic tests and it works great.
After [PR#334](https://github.com/paraglider-project/paraglider/pull/334) will be merged I'll be able to check whether it also deletes VPN connections and local network gateways (via a multicloud test).
Thanks @seankimkdy.

PR #294, opened 2024-05-09
[IBM] Add K8s Cluster support

praveingk:
Add support for kubernetes clusters.

Currently, I have a dependency on a PR(https://github.com/IBM-Cloud/container-services-go-sdk/pull/19) in container SDK, which is why for now, I have go.mod replace to a fork.

More details on the implementation :

1. By default, the worker-worker, worker-master and worker to external service(container registries) s within the private network is provisioned.
2. Pods running on the worker nodes can communicate with the service endpoints that are accessible to the subnet.
3. Hence, before creating a cluster, a new subnet is created for the cluster. The worker nodes are allocated to the subnet.
4. Upon creation of the cluster, the cloud provisions two security-groups, 1 for comms within cluster between the worker nodes(cluster scope), and other for communication outside (vpc scope).
5. For permitlists, we are interested in the latter security group (vpc scope). This security group is tagged and used for permitlists.
6. To ensure there are no shared subnets and security groups between the clusters, for each cluster, we create an exclusive VPC, and allocate subnets in it. Also, other resources created such as VMs, will not use this VPC, and will use a VPC that is tagged as shared.
7. The implication of the above implementation choice is that k8s clusters and other resources within the same region would still use transit gateway to connect to each other. As of now, it's not clear if this is a drawback. But, I feel this creates better isolation properties. Also, it makes the k8s network (vpc) more cohesive when we provision multi-zone, which I haven’t added support for right now.


smcclure20 (approved):
LGTM!

pkg/ibm_plugin/sdk/resource.go
```diff
@@ -0,0 +1,513 @@
...
+	if err != nil {
+		utils.Log.Println("failed to get create instance options: ", err)
+		return nil, err
+	}
+
+	clusterOptions.Name = &name
+	clusterOptions.XAuthResourceGroup = c.client.resourceGroup.ID
+	clusterOptions.Provider = core.StringPtr(clusterType)
+	clusterOptions.WorkerPool.VpcID = &vpcID
+	clusterOptions.WorkerPool.Zones[0].SubnetID = &subnetID
```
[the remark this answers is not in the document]
praveingk: The Pod subnets can be specified by the user through the json. Ideally, Pod IPs are local to the cluster, and will not be used in the permit list. The security group for POD communication and inter cluster communication is different. Hence, I did not see it being impacted. However, Will verify this, and if necessary change It to be obtained from orchestrator.

pkg/ibm_plugin/sdk/resource.go
```diff
@@ -0,0 +1,513 @@
...
+
+// GetID fetches the identifier of instance
+func (c *ResourceClusterType) GetID() string {
+	return c.ID
+}
+
+// GetVPC returns the VPC reference of the endpoint gateway of the cluster
+func (c *ResourceClusterType) GetVPC() (*vpcv1.VPCReference, error) {
+	// A Cluster would have a security group with prefix of 'kube-',
+	// We infer the VPC of the cluster using the VPC of this security group
```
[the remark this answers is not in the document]
praveingk: While the REST API had a way to get the VPC of a cluster, the Go SDK dint have that support. Once I add add its support in ibm-cloud/container-services-go-sdk, I can change it to retrieving it directly from the getCluster

praveingk:
@seankimkdy @divega  Can you check and approve the PR, since I had to update the branch after @smcclure20 approval

pkg/ibm_plugin/sdk/tagging.go
```diff
@@ -34,26 +34,27 @@ func (c *CloudClient) attachTag(CRN *string, tags []string) error {
...
-		utils.Log.Printf("Tagging attempt %v: Response: %+v", attempt, latestResponse)
+		utils.Log.Printf("Tagging attempt %v: Error: %v", attempt, err)
 		if !*result.Results[0].IsError {
 			return nil
 		}
 		// sleep to avoid busy waiting
 		time.Sleep(5 * time.Second)
 	}
-	return fmt.Errorf("Failed to tag resource with response:\n %+v", latestResponse)
+	return fmt.Errorf("failed to tag resource : %v", err)
```
seankimkdy, line 57: I believe you're trying to return an error at the end if all the attempts fail, but `err` gets redeclared within the scope of the for loop because of the `:=`. So line 57 actually would just have a blank `err`.

Correct me if I'm wrong though.

pkg/ibm_plugin/sdk/vpc.go
```diff
@@ -26,12 +26,18 @@ import (
 
 const vpcType = "vpc"
 
+const SharedVPC = "shared"
```
seankimkdy, line 29: Does this need to be exported?

pkg/ibm_plugin/sdk/resource.go
```diff
@@ -0,0 +1,513 @@
...
+
+	return nil, fmt.Errorf("failed to unmarshal resource description:%+v", err)
+
+}
+
+// GetResourceHandlerFromID gets the resource handler from the resource ID/URI
+func (c *CloudClient) GetResourceHandlerFromID(deploymentID string) (ResourceIntf, error) {
+	parts := strings.Split(deploymentID, "/")
+
+	if len(parts) >= 5 {
```
seankimkdy, line 503: Is this better than checking for exact length of 6?

seankimkdy (changes requested):

pkg/ibm_plugin/sdk/vpc.go
praveingk: Yes, it us used in plugin.go

seankimkdy: But they're both in the ibm package so i don't think it needs to be exported right?

praveingk: It still needs to be exported since it's in the context of ibm_plugin/sdk . 
`sdk.sharedVPC` doesn't seem to work, neither does sharedVPC..

seankimkdy: Got it no worries then

pkg/ibm_plugin/sdk/resource.go
praveingk: Since I am checking parts[5], dint want to check the exact length of 6, and checking >=5 seems more readable :)

pkg/ibm_plugin/sdk/tagging.go
praveingk: Ah, good catch..Will update

pkg/ibm_plugin/sdk/vpc.go
seankimkdy: I raised an issue related to this #318. Let's not fix it in this PR but just wanted to let you know.

pkg/ibm_plugin/sdk/tagging.go
praveingk: Done in https://github.com/paraglider-project/paraglider/pull/294/commits/d4c44242aee720ccd3a8478c4c88be751fd3e9c8

seankimkdy (approved):

PR #331, opened 2024-05-21
[GCP] Clean up URL usage and terminology

### LINKED ISSUE #164: [GCP] Consolidate URL/URI naming to URL

seankimkdy:
Resolves #164 

- Decided on using URL within `pkg/gcp/` since GCP docs tend to use URL over URI
- Make all `get___Url` functions return a fully qualified URL (i.e., with `https://www.googleapis.com/compute/v1/` prefix) since only some GCP requests allow partial URLs but all allow fully qualified ones


pkg/gcp/plugin_utils.go
```diff
@@ -290,16 +290,24 @@ func getRegionFromZone(zone string) string {
...
+func parseUrl(url string) map[string]string {
+	path := strings.TrimPrefix(url, computeUrlPrefix)
+	path = strings.TrimPrefix(path, containerUrlPrefix)
+	parsedUrl := map[string]string{}
 	pathComponents := strings.Split(path, "/")
-	for i := 0; i < len(pathComponents)-1; i += 2 {
-		parsedURL[pathComponents[i]] = pathComponents[i+1]
+	for i := 0; i < len(pathComponents)-1; {
+		if pathComponents[i] == "global" {
+			// Global resources only have a "global" specification withput a trailing value
```
smcclure20, line 302: Nit: without

pkg/gcp/plugin_utils.go
```diff
@@ -290,16 +290,24 @@ func getRegionFromZone(zone string) string {
...
 
-// Parses GCP compute URL for desired fields
-func parseGCPURL(url string) map[string]string {
-	path := strings.TrimPrefix(url, computeURLPrefix)
-	path = strings.TrimPrefix(path, containerURLPrefix)
-	parsedURL := map[string]string{}
+// parseUrl parses fully qualified and partial GCP resource URLs into a map of URL components (e.g., region)
+// and their corresponding values (e.g., us-west1).
+func parseUrl(url string) map[string]string {
+	path := strings.TrimPrefix(url, computeUrlPrefix)
```
smcclure20, line 296: Nit: add we add more resources, might be good to have a list of supported ones and iterate through it instead of having to add to this every time we add one

smcclure20 (approved):
LGTM!

pkg/gcp/plugin_utils.go
seankimkdy: My left hand is still improving

smcclure20 (approved):

PR #343, opened 2024-05-22
Add support to build docker image

praveingk:
Docker images gets published in : 
https://github.com/orgs/paraglider-project/packages/container/package/paraglider

Next step would be add releases

Note: For now, I added config.yaml (as config argument for glided startup) which can be provided when executing the docker. 
`docker run -v /path/to/your/config.yaml:/config.yaml ghcr.io/paraglider-project/paraglider:latest`

However, we should see if we can simplify this further.


smcclure20 (approved):
Thanks for doing this!

PR #352, opened 2024-05-30
Add support for CLI namespace set.

aviweit:
Currently, Paraglider namespaces are not being supported from the CLI (only 'default' can be used).

It seems that `glide namespace set` has no effect since it sets a local variable in the client itself which is stateless.

In this PR, `glide namespace set` updates the active namespace in the controller side and other CLIs query first the namespace from there and use it to create the resource and rule.

It seems to work well in my local environment. I can create resources in multiple namespaces which results in the creation of multiple VPCs.

The PR is not fully completed: few tests and lint fails.

Can you please review it?

Thanks.

### DIFF
--- a/pkg/orchestrator/orchestrator.go
+++ b/pkg/orchestrator/orchestrator.go
@@ -60,6 +60,7 @@ const (
 	DeleteTagURL             string = "/tags/:tag"
 	DeleteTagMemberURL       string = "/tags/:tag/members/:member"
 	ListNamespacesURL        string = "/namespaces"
+	SetNamespaceURL          string = "/namespaces/:namespace"
 )
 
 type Warning struct {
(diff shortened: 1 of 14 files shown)


pkg/orchestrator/orchestrator.go
```diff
@@ -1196,6 +1198,20 @@ func (s *ControllerServer) deleteTagMember(c *gin.Context) {
 	c.JSON(http.StatusOK, gin.H{})
 }
 
+// Set namespace
+func (s *ControllerServer) setNamespace(c *gin.Context) {
+	namespace := c.Param("namespace")
+	s.namespace = namespace
```
aviweit, line 1204: Should the active namespace be persisted?

praveingk (changes requested):
@aviweit Thanks for checking this, and working on it. I am skeptical about maintaining an active namespace in the orchestrator, especially when the orchestrator works across namespace with multiple clients resulting in race conditions. This is the reason, all resource ID have a namespace component. Ideally, the cliSettings could be persisted in the client side (e.g. by json marshaling cliSettings to a file), rather than being propagating it to the orchestrator to keep it simple, and each client independent. @smcclure20 @seankimkdy Let me know what you think.

aviweit:
Thanks for clarifying this. I am closing this PR.

praveingk:
@aviweit : Will you be opening another PR with persistency support for namespace in cliSettings? 
cc : @smcclure20

PR #356, opened 2024-06-03
Add release workflow

praveingk:


praveingk:
I am still testing in my local env.

praveingk:
I have tested it in my local env using act, and it works.

seankimkdy (commented):
@praveingk This is great! A few concerns before I stamp.

1. Seems like this workflow is triggered on a tag push to the main branch. Do we need any [tag protection rules](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules) to ensure only maintainers can push tags to the main branch?
2. There's also concerns of [tags on other branches triggering this type of workflow](https://github.com/orgs/community/discussions/13226). Can we double check and make sure this doesn't happen or ensure it's not really an issue?
3. Have you tested the actual binaries that have been created?

praveingk:
> @praveingk This is great! A few concerns before I stamp.
> 
> 1. Seems like this workflow is triggered on a tag push to the main branch. Do we need any [tag protection rules](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules) to ensure only maintainers can push tags to the main branch?
> 2. There's also concerns of [tags on other branches triggering this type of workflow](https://github.com/orgs/community/discussions/13226). Can we double check and make sure this doesn't happen or ensure it's not really an issue?
> 3. Have you tested the actual binaries that have been created?



> @praveingk This is great! A few concerns before I stamp.
> 
> 1. Seems like this workflow is triggered on a tag push to the main branch. Do we need any [tag protection rules](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules) to ensure only maintainers can push tags to the main branch?

Yes, we will need to check and add protection rules if its not already present
> 2. There's also concerns of [tags on other branches triggering this type of workflow](https://github.com/orgs/community/discussions/13226). Can we double check and make sure this doesn't happen or ensure it's not really an issue?

I have changed the "on" action to push on main branch. I am unable to test it out in my local env though.
> 3. Have you tested the actual binaries that have been created?

I had used act to test the artifacts, and the binaries worked.

smcclure20:
@seankimkdy @praveingk I created a tag protection rule that (for now) requires maintainers to push, modify, or delete all tags. We can weaken this later if needed.

smcclure20 (approved):

PR #362, opened 2024-06-04
Persist CLI-local settings

smcclure20:
The CLI was not persisting its settings (they were only stored in memory, which does not persist between commands). This PR fixes this by writing settings to file.


divega (approved):
Curious if the change from "settings" to "config" is out of personal preference or if there is more to it. Also like the idea is .paraglider/settings.json form the other PR.

internal/cli/glide/config/config.go
```diff
@@ -0,0 +1,92 @@
+/*
+Copyright 2023 The Paraglider Authors.
```
praveingk, line 2: Should we have 2024 in new files?

praveingk (approved):
Looks good

PR #371, opened 2024-06-06
Integration test for multiple region Intra-namespace

### LINKED ISSUE #235: [Azure] Add multi-region, intra-namespace integration test

J-467:
- Added an integration test for VMs within the same namespace and in multiple regions on Azure
- Removed a duplicate error check in `plugin.go`
- closes #235

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,5 @@ docs/build/
 
 # Python
 .venv/
+
+config/paraglider_config.yaml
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -45,3 +45,5 @@ docs/build/
 
 # Python
 .venv/
+
+config/paraglider_config.yaml
```
seankimkdy, line 49: Where did this `config/` folder come from?

J-467: My bad. This is a local folder I added for glided. Should have removed it before PR'ing. Sending a fix

PR #334, opened 2024-05-22
[IBM] VPN enablement and integration with Azure

Cohen-J-Omer:
This PR provides VPN support for the IBM cloud plugin, including necessary adjustments to the controller and Azure plugin, to support BGP disabled VPN connections.
This PR is an extension of [PR#179](https://github.com/paraglider-project/paraglider/pull/179), that was automatically closed due to a project name change.
Future PR will make the necessary adjustments to the IBM and GCP plugins, supporting IBM-GCP connectivity.   

Discussion Topics:
- Integration test file still contains the `ibm` build tag, as Github's workflow wouldn't be able to run it without IBM account credentials. When the issue is handled, I'll move the IBM-Azure test to `multicloud_test.go`.
- I aim to add an IBM `RunPingConnectivityCheck()`, equivalent to Azure's. This method will require ssh login (a key is produced by `auth.go`), as I'm unaware of a network watcher equivalent on IBM.

PR #386, opened 2024-06-12
Moving general functions from pkg/ibm to pkg/utils

J-467:
Moving the CIDR comparison functions from IBMs utils to the general utils folder, so it's available for all the clouds

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package log
+package utils
 
 import (
 	"fmt"
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -19,6 +19,7 @@ package log
 import (
 	"fmt"
 	"log"
+	"net"
 	"net/netip"
 	"os"
 	"strings"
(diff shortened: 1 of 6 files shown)


pkg/utils/utils_test.go
```diff
@@ -16,7 +16,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package ibm
+package log
```
seankimkdy, line 19:
```suggestion
package utils
```

Not sure how this became `package log`, but could you also change this in `pkg/utils/utils.go`?

seankimkdy (changes requested):

pkg/utils/utils.go
```diff
@@ -172,3 +173,44 @@ func GetNumVpnConnections(cloud1, cloud2 string) int {
...
+// i.e. they don't overlap.
+func DoCIDROverlap(cidr1, cidr2 string) (bool, error) {
+	netCIDR1, err := netip.ParsePrefix(cidr1)
+	if err != nil {
+		return true, err
+	}
+	netCIDR2, err := netip.ParsePrefix(cidr2)
+	if err != nil {
+		return true, err
+	}
```
seankimkdy, line 187: Why does it make sense to return true if we fail to parse either IP?

J-467: I don't think there's a particular reason/advantage to return either true or false in this case because the idea is to check for no errors before using result value. However, I'd understand true in the sense that this function might be used in the context of ensuring that some ip ranges don't overlap (so true, signifying an overlap might be good?)

pkg/utils/utils_test.go
J-467: Updated

Cohen-J-Omer (commented):
Hey @J-467, looks good and the change behind the PR makes sense, but can we postpone merging it until [PR#334](https://github.com/paraglider-project/paraglider/pull/334) is merged?

J-467:
> Hey @J-467, looks good and the change behind the PR makes sense, but can we postpone merging it until [PR#334](https://github.com/paraglider-project/paraglider/pull/334) is merged?

@Cohen-J-Omer Yup, we can

J-467:
Created a different PR for this. #411

PR #387, opened 2024-06-12
[Azure] Compliance-check for functions for existing resources

### LINKED ISSUE #385: [Azure] Determine existing resource compliance with Paraglider
Determine existing resource/network compliance with Paraglider
  - [x] Validate that resource is in the provided resource group
  - [x] Validate that resource group conforms to the paraglider deployment namespace
  - [x] Ensure VNet address isn't overlapping within the deployment
  - [x] Ensure that permit list of the VNet has a deny all rule with the lowest priority. Create one if none exists.
  - [x] Ensure that permit list of the VNet has a deny all rule with the lowest priority (i.e. highest priority number). Create one if none exists.
  - [x] Ensure that there's no deny rule with a higher priority than an allow rule
  - [x] Any deny rule must be a deny all rule. All rule verifications disregard the default rules created by Azure with priority > 65000.
  - [ ] Add unit test for the overarching function `DoesVnetOverlapWithParaglider`

_Originally posted by @J-467 in https://github.com/paraglider-project/paraglider/issues/170#issuecomment-2161210207_

J-467:
- Adds functions and tests to determine the compliance of existing Azure resources with paraglider for attachment
- Criteria for compliance:
  - [x] Validate that resource is in the provided resource group
  - [x] Validate that resource group conforms to the paraglider deployment namespace
  - [x] Ensure VNet address isn't overlapping within the deployment
  - [x] Ensure that permit list of the VNet has a deny all rule with the lowest priority (i.e. highest priority number). Create one if none exists.
  - [x] Ensure that there's no deny rule with a higher priority than an allow rule
  - [x] Any deny rule must be a deny all rule. All rule verifications disregard the default rules created by Azure with priority > 65000.
- Closes #385


pkg/azure/resources.go
```diff
@@ -148,6 +172,40 @@ func ReadAndProvisionResource(ctx context.Context, resource *paragliderpb.Create
 	return handler.readAndProvisionResource(ctx, resource, subnet, resourceInfo, sdkHandler, additionalAddressSpaces)
 }
 
+func DoesVnetOverlapWithParaglider(ctx context.Context, handler *AzureSDKHandler, vnetName string, server *azurePluginServer) (bool, error) {
```
J-467, line 175: @smcclure20  I'm thinking this function belongs to a different file. (we're passing in the azurePluginServer here). Thoughts? Or is it fine here?

pkg/utils/utils.go
```diff
@@ -172,3 +172,21 @@ func GetNumVpnConnections(cloud1, cloud2 string) int {
 	}
 	return 1
 }
+
+// DoCIDROverlap returns false if cidr blocks don't share a single ip,
+// i.e. they don't overlap.
+func DoesCIDROverlap(cidr1, cidr2 string) (bool, error) {
```
J-467, line 178: I'll delete this function before merging. Created another PR that moves a similar function from pkg/ibm to pkg/utils so it's available to all clouds

dump.rdb
J-467, line 1: I'll delete this file

J-467 (commented):
Writing tests for these functions. Will push soon

pkg/azure/resources.go
```diff
@@ -55,6 +55,10 @@ func getClusterUri(subscriptionId string, resourceGroupName string, clusterName
 	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s/providers/%s/%s", subscriptionId, resourceGroupName, managedClusterTypeName, clusterName)
 }
 
+func getDeploymentUri(subscriptionId string, resourceGroupName string) string {
+	return fmt.Sprintf("/subscriptions/%s/resourceGroups/%s", subscriptionId, resourceGroupName)
+}
+
```
smcclure20, line 61: This might be better placed in naming.go?

pkg/azure/resources.go
```diff
@@ -139,6 +143,26 @@ func GetResourceInfoFromResourceDesc(ctx context.Context, resource *paragliderpb
 	return handler.getResourceInfoFromDescription(ctx, resource)
 }
 
+// Verify that resource is supported, and return the resource ID
+func GetResourceIDFromName(ctx context.Context, handler *AzureSDKHandler, resourceGroup string, resourceName string) (string, error) {
```
smcclure20, line 147: Based on the name, I would expect this to just do what `getVmUri` does. Maybe rename to something like `ConfirmResourceExists`

pkg/azure/resources.go
smcclure20: Hm, yeah this might be better somewhere else. Maybe just in utils?

smcclure20 (changes requested):
Looks good! Though, this only checks if the vnet address space overlaps. We will also need to check the NSG. And, as a potential later step, make sure the vnet is not peered with any other vnets that overlap. We can discuss the last one more later.

pkg/azure/resources.go
J-467: Sounds good. Updated

pkg/azure/rule_utils.go
```diff
@@ -0,0 +1,189 @@
...
+	if lowestRule == nil {
+		if reservedPriorities[maxPriority] != nil {
+			// The max priority number should not be associated to an allow rule
+			return -1, fmt.Errorf("Allow Rule at lowest priority(%d). Must be a deny all rule", maxPriority)
+		}
+
+		return maxPriority, fmt.Errorf("No deny rule present")
+	}
+
+	// If not a deny all rule, return priority to create a deny all rule
```
smcclure20, line 114: Either way, we will need the new deny-all to override the default azure rules which are usually of the form:
{Priority: 65500, Name: Allow-all-vnet}
{Priority: 65501, Name: deny-all}
So no matter what, we want to put a deny-all rule with higher (lower-value) priority than these rules.

pkg/azure/rule_utils.go
```diff
@@ -0,0 +1,189 @@
...
+	if !isDenyAllRule(lowestRule) {
+		lastPriority := getPriority(reservedPriorities, minPriority, maxPriority, false)
+		return lastPriority, fmt.Errorf("Deny Rule not at lowest priority(%d) is not a Deny all rule", lowestDenyPriorityNum)
+	}
+
+	return lowestDenyPriorityNum, nil
```
smcclure20, line 120: Overall, the only reason we would need to check for a deny-all is if they already have one that is higher priority (lower value) than the default azure allow vnet rules. In that case, we can use that rule as our deny-all.
Otherwise, the requirement for deny rules (not specifically deny-all) is that there shouldn't be any.

So to summarize the requirements:
* No deny rules that are not deny-all
* The only acceptable deny-all rules are (1) the lowest priority in the paraglider range (100-4096) or (2) just above the default Azure rules (<65500)

We will override (1) and the default azure rules directly above it that allow things like the local vnet traffic. And for (2), we can use it as our inserted deny-all if we want. Or we can just say that (2) is not allowed (to make things simple) and then insert our deny-all for everything. 

So for (2), let's do the simple option: it doesn't cause a failure, but we still override it with a deny-all at 4096.

pkg/azure/priority_utils.go
```diff
@@ -17,40 +17,46 @@ limitations under the License.
...
 }
 
-// getPriority returns the next available priority that is not used by other rules
-func getPriority(reservedPriorities map[int32]bool, start int32, end int32) int32 {
-	var i int32
-	for i = start; i < end; i++ {
-		if !reservedPriorities[i] {
-			reservedPriorities[i] = true
+// getPriority returns the next available priority number that is not used by other rules
+func getPriority(reservedPriorities map[int32]*armnetwork.SecurityRule, start int32, end int32, ascendingSearch bool) int32 {
```
smcclure20, line 47: Nit: rename to something a bit more accurate like "getNextAvailablePriority"

pkg/azure/priority_utils.go
```diff
@@ -17,40 +17,46 @@ limitations under the License.
...
+	if !ascendingSearch {
+		start, end = -end, -start
+	}
+
+	var idx int32
+	for i := start; i < end; i++ {
```
smcclure20, line 53: Not sure if this is better, but you could just set an `increment` variable to +1 or -1 depending on the direction of the search and then you don't have to mess with absolute values. 

Example:
```
       if !ascendingSearch {
		start, end = end, start
                 increment = -1
	}

	for i := start; i < end; i+increment {
```

pkg/azure/priority_utils.go
smcclure20, line 47: Also, this interface is a little awkward I think. Maybe even just calling `ascendingSearch` something like `lowestAvailable` or something so that its less implementation-specific. And either way, I think you should add to the docstring with information about the function arguments.

smcclure20 (changes requested):
This is great! Thanks for all the work! I just left mostly some nits and a few changes with the compliance logic.

smcclure20 (approved):
Looks good!

PR #391, opened 2024-06-14
Fix release trigger on PR Merge

praveingk:
Currently, a release is triggered upon PR merge. I realized it's because of `branch:main` in push, which doesn't filter out only tag pushes, but applies to all pushes (PR merge) to the main branch. 
I referred to bunch of other projects, which don't have this condition set. Hence, we should be ok with not having `branch:main` condition.


smcclure20 (approved):
Thanks for catching this!

PR #392, opened 2024-06-14
Enable Permit List Rule Operations on Tags

### LINKED ISSUE #117: Implement resolving tags to URIs for applying functions
As of now, tags can only be used as a list of IPs for permit list rules to reference. We should support referencing tags in requests to change permit lists so that the user can perform bulk operations.

smcclure20:
Enables users to apply rules (or delete rules) on tags. Before, users could only apply a rule on one resource at a time. Now, if a tag maps to many resources, this can be leveraged to apply a rule in one command to many resources.

Resolves #117

### DIFF
--- a/pkg/orchestrator/orchestrator.go
+++ b/pkg/orchestrator/orchestrator.go
@@ -397,6 +397,8 @@ func (s *ControllerServer) permitListRuleAddTag(c *gin.Context) {
 		return
 	}
 
+	initializedClientConns := make(map[string]*grpc.ClientConn)
+
 	// Add rule to each URI in the resolved tag
 	for _, mapping := range resolvedTag.Tags {
 		// Get the cloud and namespace from the tag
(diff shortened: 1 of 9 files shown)


pkg/orchestrator/orchestrator.go
```diff
@@ -361,6 +370,124 @@ func (s *ControllerServer) permitListRuleAdd(c *gin.Context) {
...
+			return
+		}
+
+		// Create connection to cloud plugin
+		cloudClient, ok := s.pluginAddresses[cloud]
+		if !ok {
+			c.AbortWithStatusJSON(400, createErrorResponse("invalid cloud name"))
+			return
+		}
+		conn, err := grpc.NewClient(cloudClient, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
praveingk, line 415: The grpc new client and initialization could be moved out of the loop

praveingk (approved):
Looks good, Just left a minor comment on moving the grpc init outside the loop.

pkg/orchestrator/orchestrator.go
smcclure20: But the tags may point to resources in different clouds, so we can't quite do that. I guess I could keep track of which clients have been initialized in the loop and use them if they have - not sure if that is worth it though. Thoughts?

praveingk (approved):
Looks good.

PR #396, opened 2024-06-20
Changing Invisinets prefix to Paraglider prefix in repo

### LINKED ISSUE #291: Rename all environment variables from `INVISINETS` prefix to `PARAGLIDER` prefix
We have several environment variables that all start with `INVISINETS`. We should rename them.

J-467:
closes #291


smcclure20 (approved):
Thanks for updating this!

PR #402, opened 2024-06-26
[Azure] Add vnet to paraglider

J-467:
- `AttachResource` function to add a resource with a specified resourceID and namespace to the paraglider namespace by creating a peering to gateway vnet.
- Added unit tests for the AttachResource function
- Does not rename the vnet, but would tag the resource (to be implemented in another PR)
- Refactor some existing tests and functions to enhance reusability

#AttachResourceAzure


pkg/azure/plugin.go
```diff
@@ -864,3 +813,41 @@ func Setup(port int, orchestratorServerAddr string) *azurePluginServer {
 	}()
 	return azureServer
 }
+
+// Add an existing Azure resource to a paraglider deployment
+func (s *azurePluginServer) AttachResource(ctx context.Context, attachResourceReq *paragliderpb.AttachResourceRequest) (*paragliderpb.AttachResourceResponse, error) {
```
smcclure20, line 818: Nit: I think the `Setup()` function should be last in the file

pkg/azure/sdk_handler.go
```diff
@@ -290,8 +290,8 @@ func (h *AzureSDKHandler) DeleteSecurityRule(ctx context.Context, nsgName string
 	return nil
 }
 
-// GetVnetAddressSpaces returns a map of location to address space for all virtual networks (VNets) with a given prefix.
-func (h *AzureSDKHandler) GetVNetsAddressSpaces(ctx context.Context, prefix string) (map[string][]string, error) {
+// Returns a map of location to address space for all virtual networks (VNets) with a given prefix.
+func (h *AzureSDKHandler) GetVnetsWithMatchingPrefixAddressSpaces(ctx context.Context, prefix string) (map[string][]string, error) {
```
smcclure20, line 294: This new name confuses me -- why the change?

smcclure20: Ah, I see. Maybe instead: `GetAllVnetAddressSpaces`?

pkg/azure/test_utils.go
```diff
@@ -36,6 +36,8 @@ import (
 )
 
 const (
+	fakeNsgName                              = "test-nsg-name"
+	fakeNsgID                                = "a/b/" + fakeNsgName
```
smcclure20, line 40: Maybe make this properly formatted like `uriPrefix + "Microsoft.Network/networkSecurityGroups/" + name`? In fact, isn't this redundant with `validSecurityGroupID` below?

pkg/azure/test_utils.go
```diff
@@ -428,13 +449,32 @@ func getFakeSubnet() *armnetwork.Subnet {
...
+				{
+					Name: to.Ptr(validSubnetName),
+					ID:   to.Ptr(validParagliderSubnetId),
+				},
+			},
+		},
+	}
+}
+
+func getFakeVirtualNetworkWithUnusedAddressSpace() *armnetwork.VirtualNetwork {
```
smcclure20, line 469: I assume this is for creating fake vnets for resources to attach? I think a better name might make that more clear. How about something like `getFakeUnattachedVirtualNetwork` (and then you could follow this naming convention for other relevant things if necessary.

smcclure20 (changes requested):
Just a few minor comments! 

For the future though, maybe it would be good to separate refactoring/reorganizing from new functionality as much as possible in PRs just so they are a little easier to read.

pkg/azure/test_utils.go
J-467: actually, it is

pkg/azure/test_utils.go
J-467: Sounds good. Will fix this

pkg/azure/sdk_handler.go
J-467: I need to work on choosing better variable and function names. I'll go with this and add function doc to it.

pkg/azure/plugin.go
J-467: Done

J-467:
> Just a few minor comments!
> 
> For the future though, maybe it would be good to separate refactoring/reorganizing from new functionality as much as possible in PRs just so they are a little easier to read.

Well noted! Thanks Sarah

pkg/azure/plugin.go
```diff
@@ -851,3 +830,24 @@ func (s *azurePluginServer) AttachResource(ctx context.Context, attachResourceRe
...
+	paragliderpb.RegisterCloudPluginServer(grpcServer, azureServer)
+	fmt.Println("Starting server on port: ", port)
+
+	go func() {
+		if err := grpcServer.Serve(lis); err != nil {
+			fmt.Println(err.Error())
+		}
+	}()
+	return azureServer
+}
```
smcclure20, line 853: Nit: newline at EOF

smcclure20 (changes requested):
Just one nit, otherwise looks great!

pkg/azure/plugin.go
J-467: Done

smcclure20 (approved):

PR #403, opened 2024-06-27
[Infra] Add approval workflow for functional tests

seankimkdy:
I added an approval workflow where running functional tests requires an approval from @paraglider-project/maintainers. This was achieved through using [environments](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment). If you head to the Settings > Environments, you should see one that called "functional-tests".

Here is a break down of the process.

1. Creating/updating a PR will trigger a check called "Functional Tests Approval". This will not run automatically unless you are an admin of this repo.
Screenshot 2024-07-03 at 04 21 49

2. Maintainers can click on the "Details" link as shown above. Here click the "Review pending deployments" to give an approval and leave any comments if desired.
Screenshot 2024-07-03 at 04 22 08

3. The "Functional Tests Approval" workflow will run. It saves the commit SHA as an artifact for the next part.

4. Upon completion of this workflow, the "Functional Tests" workflow will automatically run due to the workflow_run trigger. This is where the functional tests will run.

You might be wondering why we need this complicated two-tiered system. It has to do with the tokens that GitHub Actions uses to run workflows and how permissions change for external contributors of the repo (i.e., pull requests from forks). I've crafted a [doc](https://docs.google.com/document/d/1cL0uES9L3kKUd-dlwxWROIlXlxOovhH1E1mUpMowHJo/edit?usp=sharing) explaining this in detail. The method I implemented is based off the "Gated Approvals" section of that doc (credits to @rynowak).

> [!NOTE]
> Due to the way we run the "Functional Tests" workflow using the `on: workflow_run` trigger of "Functional Tests Approvals", the "Functional Tests" workflow will not run in the checks for this PR since functional-tests.yaml is not in the main branch yet (see https://github.com/orgs/community/discussions/25288). Once we merge this PR into main, the "Functional Tests" workflow should run, but there may be issues that require follow-up PRs.


.github/workflows/functional-tests.yaml
```diff
@@ -0,0 +1,82 @@
...
+  # gotestsum version - see: https://github.com/gotestyourself/gotestsum
+  GOTESTSUMVERSION: 1.10.0
+
+  # github contexts
+  GH_RUN_ID: ${{ github.run_id }}
+  GH_RUN_NUMBER: ${{ github.run_number }}
+  GH_RUN_ATTEMPT: ${{ github.run_attempt }}
+
+  # Cloud account fields
+  INVISINETS_GCP_PROJECT_PARENT: ${{ secrets.GCP_FOLDER_ID }}
```
smcclure20, line 23: Can the same lines be removed from build.yaml since it isn't using the clouds at all any more?

smcclure20 (approved):
Just a small (potential) nit. Thanks for working on this!

.github/workflows/functional-tests.yaml
seankimkdy: Good point

smcclure20 (approved):

PR #406, opened 2024-06-28
[Docs] Update Azure readme

J-467:
- Update Azure readme on local testing
- update the prerequisite to include the install make command

### DIFF
--- a/docs/source/developers/contributing/prerequisites.rst
+++ b/docs/source/developers/contributing/prerequisites.rst
@@ -35,8 +35,10 @@ This is the list of core dependencies to install for the most common tasks. In g
     
         $ apt install build-essential
 
-      Then install ``make``:
+    Then install ``make``:
   
+    .. code-block:: console
+
         $ apt install make
   
   * Mac:
--- a/docs/source/developers/contributing/prerequisites.rst
+++ b/docs/source/developers/contributing/prerequisites.rst
@@ -34,6 +34,10 @@ This is the list of core dependencies to install for the most common tasks. In g
     .. code-block:: console
     
         $ apt install build-essential
+
+      Then install ``make``:
+  
+        $ apt install make
   
   * Mac:
 
--- a/pkg/azure/README.md
+++ b/pkg/azure/README.md
@@ -102,9 +102,11 @@ Follow these steps to ensure consistent structure and seamless integration.
      }
      // Use handler
      ```
+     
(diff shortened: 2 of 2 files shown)


pkg/azure/README.md
```diff
@@ -84,24 +95,24 @@ Follow these steps to ensure consistent structure and seamless integration.
...
+     handler, err = s.setupAzureHandler(resourceIdInfo, req.Deployment.Namespace)
      if err != nil {
          return nil, err
      }
+     // Use handler
      ```
 
 3. Handling ARM Requests:
-   - If your new API needs to make ARM requests, check if the required functionality is already implemented in `sdk_handler.go`. If not, add a new function in the handler that wraps the Azure request. Call this new function from the `plugin` file. Any helper functionalities relevant only to Paraglider and the server can remain in `plugin`.
+   - If your new API needs to make ARM requests, check if the required functionality is already implemented in `sdk_handler.go`. If not, add a new function in the handler that wraps the Azure request. Call this new function from the `plugin` file. Ideally, helper functionalities relevant to Paraglider and the server should be outside `plugin`.
```
smcclure20, line 107: Could we say "Ideally, only the Paraglider plugin interface RPCs should be in `plugin` and helper functions should go in the appropriate file" or something like that?

smcclure20 (changes requested):
Thanks for updating this!

pkg/azure/README.md
J-467: Updated

smcclure20 (approved):

docs/source/developers/contributing/prerequisites.rst
```diff
@@ -34,6 +34,10 @@ This is the list of core dependencies to install for the most common tasks. In g
     .. code-block:: console
     
         $ apt install build-essential
+
+      Then install ``make``:
+  
+        $ apt install make
```
seankimkdy, line 40:
```suggestion
    Then install ``make``:
  
    .. code-block:: console

        $ apt install make
```

Spacing is a little tricky for RST. Right now, what you currently have looks like this

Screenshot 2024-07-04 at 00 01 07

I assume you want it to look like this.

Screenshot 2024-07-04 at 00 01 15

seankimkdy (changes requested):

seankimkdy (approved):

Issue #170, opened 2024-03-14
Idea: Allow for brownfield deployment by connecting VPCs to deployment if their addresses meet requirements


smcclure20:

divega:
Triage: this is part of the work Julian will be looking into. We need to split or keep it open because Julian will be focusing only on the Azure side. @J-467, please create detailed check list of tasks.

J-467:
- [ ] Validate that resource is in the provided resource group
- [ ] Validate that resource group conforms to the paraglider deployment namespace
- [ ] Ensure VNet address isn't overlapping within the deployment

PR #410, opened 2024-07-01
[Azure] Add Paraglider namespace tag to resource vnet when attaching

J-467:
- For resource attachment, we should add the paraglider namespace tag to the virtual network. 
- This tag would be useful for identifying which vnet in user's subscription is managed by paraglider.
- Simple refactor in `CreateParagliderVirtualNetwork` to utilize `CreateOrUpdateVirtualNetwork`
- Contributes to #170


pkg/azure/sdk_handler.go
```diff
@@ -573,22 +573,16 @@ func (h *AzureSDKHandler) CreateParagliderVirtualNetwork(ctx context.Context, lo
...
 	}
 
-	return &resp.VirtualNetwork, nil
+	return vnet, nil
 }
 
-func (h *AzureSDKHandler) CreateVirtualNetwork(ctx context.Context, name string, parameters armnetwork.VirtualNetwork) (*armnetwork.VirtualNetwork, error) {
+// Updates properties of the virtual network (vnet) if it exists. Creates a new vnet if it doesn't exist.
+func (h *AzureSDKHandler) CreateOrUpdateVirtualNetwork(ctx context.Context, name string, parameters armnetwork.VirtualNetwork) (*armnetwork.VirtualNetwork, error) {
 	h.createParagliderNamespaceTag(&parameters.Tags)
```
smcclure20, line 586: Hm, I like the separation with the fact that `CreateParagliderVirtualNetwork` just calls `CreateorUpdateVirtualNetwork` with the parameters to make a Paraglider vnet, but then this seems like a weird place for this line to be (since it is paraglider specific. So, maybe pull this out and call it separately?

smcclure20 (changes requested):
Just one small comment about where to put Paraglider-specific stuff. Otherwise looks great!

pkg/azure/sdk_handler.go
J-467: I understand this

smcclure20 (approved):

smcclure20 (approved):

PR #411, opened 2024-07-01
Refactor: Moving utils from ibm to general

J-467:
Moving the CIDR comparison functions from IBMs utils to the general utils folder, so it's available for all the clouds


praveingk (approved):
Looks good

Cohen-J-Omer (approved):

PR #416, opened 2024-07-03
[Azure] Update GetAndCheckResourceState

J-467:
- `GetAndCheckResourceState` checks that a resource is valid and is in the specified namespace
- Currently, we only check prefix of resource vnet name to determine if it's in the paraglider namespace
- For attach resource, the vnet name wouldn't necessarily conform to the Paraglider naming requirements
- Updating GetAndCheckResourceState to check the ARM tags of the vnet for the namespace if the vnet name doesn't conform to the paraglider naming convention
- Includes tests to validate the added tag check
- Contributes to #170


pkg/azure/test_utils.go
```diff
@@ -641,6 +697,30 @@ func getFakeNIC() *armnetwork.Interface {
 	}
 }
 
+func getFakeUnattachedNIC() *armnetwork.Interface {
```
smcclure20, line 700: I don't see where this is used. With getFakeParagliderInterface() and getFakeInterface(), what additional purpose does this serve? But maybe my confusion is just from the basing on your other branches.

smcclure20 (approved):
I'm just a little confused about the need for three different face NICs, but otherwise looks good!

pkg/azure/test_utils.go
J-467: I mistakenly duplicated a few functions. Fixed it now

smcclure20 (approved):

Issue #394, opened 2024-06-17
Initial Support for Multitenant Services in the Plugins


smcclure20:
This is an initial step towards full support (#344), specifically only targeting the necessary plugin modifications.

**RPC Changes**
We will modify paragliderpb.proto to expose a new RPC which "attaches" resources to the paraglider deployment. The request will require a paraglider deployment, a resource name, and a URI for the resource. This is semantically different than CreateResource given that the service the user is aiming to add to Paraglider already exists.

**Plugin Implementation**
The plugin implementation of AttachResource will take the provided URI and determine if it is a multitenant service. If so, it will provision a private endpoint in the paraglider VPC in the region specified by the URI and attach it to the service. It will then create a security group (in alignment with Paraglider CreateResource expectations) and attach it to the private endpoint.

PR #417, opened 2024-07-04
[GCP] Private Endpoint Support

smcclure20:
[**Update**] Closed #435 and merged it into this PR. Originally, this PR used the URI of the service attachment for later references of the PSC (ex, adding/removing rules on it). However, since Google services do not have a service attachment, I changed the returned URI to be the forwarding rule that implements the PSC. Users shouldn't even have to use this URI since they can refer to the connection via its name and the tag service will map that to the URI.

Completes the GCP portion of #394 

**Anticipated UI**
For a minimal change to the existing user interface and in alignment with our discussions on the API, this does not require any front-end changes. Instead, users should send requests for private service connections (PSCs) by sending a CreateResource request with the description having one field: "url" which provides the URL of the service attachment to connect to (https://cloud.google.com/vpc/docs/configure-private-service-connect-services)

We can revisit this as needed as we implement for other clouds.

**Also in this PR**
- Some cleanup of `resources.go` 
- Creation of a clients struct (like Azure) so that functions do not have to have arguments for each client that may be necessary (this became especially bad as each supported resource requires different -- and potentially many -- clients, but they all have to match the same interface)


pkg/gcp/resources.go
```diff
@@ -330,38 +392,63 @@ func (r *gcpInstance) fromResourceDecription(resourceDesc []byte) (*computepb.In
 }
 
 // GCP cluster resource handler
-type gcpGKE struct {
-	gcpResourceHandler[container.ClusterManagerClient]
+type clusterHandler struct {
+	GCPResourceHandler
+	client          *container.ClusterManagerClient
+	firewallsClient *compute.FirewallsClient
```
seankimkdy, line 398: Do you think this is better than just embedding the entire `*GCPClients`? A benefit I can see with embedding `*GCPClients` is if `clusterHandler` ever needs another client, there will be minimal code changes.

pkg/gcp/resources.go
```diff
@@ -179,67 +220,87 @@ func IsValidResource(ctx context.Context, resource *paragliderpb.CreateResourceR
...
 }
 
-// Generic GCP resource handler
-type gcpResourceHandler[T supportedGCPResourceClient] struct {
-	iGCPResourceHandler
-	client *T
+// GCP instance resource handler
+type instanceHandler struct {
+	GCPResourceHandler
+	client *compute.InstancesClient
```
seankimkdy, line 258: Same comment about embedding `*GCPClients` `clusterHandler`.

pkg/gcp/resources.go
```diff
@@ -488,12 +575,150 @@ func (r *gcpGKE) fromResourceDecription(resourceDesc []byte) (*containerpb.Creat
...
+// Get the resource information about a service attachment
+func (r *privateServiceHandler) getResourceInfo(ctx context.Context, resource *paragliderpb.CreateResourceRequest) (*resourceInfo, error) {
+	description := &ServiceAttachmentDescription{}
+	err := json.Unmarshal(resource.Description, description)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse resource description: %w", err)
+	}
+	urlParts := strings.Split(description.Url, "/")
+	serviceName := urlParts[len(urlParts)-1]
+	region := urlParts[len(urlParts)-3]
```
seankimkdy, line 628: Thoughts on using `parseUrl` in `plugin_utils.go`? Would be useful as some users might pass in full vs. partial url.

pkg/gcp/resources.go
```diff
@@ -488,12 +575,150 @@ func (r *gcpGKE) fromResourceDecription(resourceDesc []byte) (*containerpb.Creat
...
+	// Reserve an IP address to be the endpoint
+	addressName := getAddressName(resourceInfo.Name)
+	addrRequest := computepb.InsertAddressRequest{
+		Project: resourceInfo.Project,
+		Region:  resourceInfo.Region,
+		AddressResource: &computepb.Address{
+			Name:        &addressName,
+			Subnetwork:  proto.String(getSubnetworkUrl(resourceInfo.Project, resourceInfo.Region, subnetName)),
+			AddressType: proto.String("INTERNAL"),
+			IpVersion:   proto.String("IPV4"),
```
seankimkdy, line 681: Let's make these `const`.

pkg/gcp/resources.go
```diff
@@ -488,12 +575,150 @@ func (r *gcpGKE) fromResourceDecription(resourceDesc []byte) (*containerpb.Creat
...
+	if err != nil {
+		return "", "", fmt.Errorf("unable to get address: %w", err)
+	}
+
+	// Create a forwarding rule for the endpoint
+	forwardingRuleRequest := computepb.InsertForwardingRuleRequest{
+		Project: resourceInfo.Project,
+		Region:  resourceInfo.Region,
+		ForwardingRuleResource: &computepb.ForwardingRule{
+			Name:      proto.String("forwarding-rule-" + resourceInfo.Name),
```
seankimkdy, line 708: Do we need to add the paraglider prefix here? On the same note, is it beneficial to make this into a function like we have for other resource names?

seankimkdy (changes requested):
Thanks for working on this @smcclure20! Mostly looks good although I'm very new to services in general so it may be helpful for someone like @praveingk or @divega to look at this at a higher level. I left comments mostly about code.

Is it possible to have an integration test for this? Would definitely be helpful to see a barebones example even if it's contrived. But also understand if this is like our k8s example where it's hard to implement an integration test.

praveingk (changes requested):

praveingk:
Thanks for adding this support. I have reviewed and posted some suggestions.

pkg/gcp/clients.go
```diff
@@ -0,0 +1,211 @@
...
+	routersClient             *compute.RoutersClient
+	vpnGatewaysClient         *compute.VpnGatewaysClient
+	vpnTunnelsClient          *compute.VpnTunnelsClient
+	externalVpnGatewaysClient *compute.ExternalVpnGatewaysClient
+	addressesClient           *compute.AddressesClient
+	forwardingClient          *compute.ForwardingRulesClient
+	serviceAttachmentClient   *compute.ServiceAttachmentsClient
+}
+
+func (c *GCPClients) GetInstancesClient(ctx context.Context) (*compute.InstancesClient, error) {
```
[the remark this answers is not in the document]
J-467: +1 for `GetOrCreate`

pkg/gcp/resources.go
```diff
@@ -33,9 +33,10 @@ import (
 )
 
 const (
-	clusterTypeName   = "cluster"
-	instanceTypeName  = "instance"
-	clusterNameFormat = "projects/%s/locations/%s/clusters/%s"
+	clusterTypeName           = "cluster"
+	instanceTypeName          = "instance"
+	serviceAttachmentTypeName = "serviceAttachment"
```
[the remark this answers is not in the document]
smcclure20: Ah, that's fair. The main reason I kept it as `serviceAttachment` is due to how it is used in like 142. We have to look at the URL for its type (`serviceAttachment`) and I thought it was best to be consistent. I could see an argument for renaming the handler though as you mention below.

pkg/gcp/resources.go
```diff
@@ -110,62 +139,74 @@ func parseResourceUrl(resourceUrl string) (*resourceInfo, error) {
...
-		handler.client = instancesClient
-		return handler, nil
+		handler = &instanceHandler{}
 	} else if resourceType == clusterTypeName {
-		handler := &gcpGKE{}
-		handler.client = clusterClient
-		return handler, nil
+		handler = &clusterHandler{}
+	} else if resourceType == serviceAttachmentTypeName {
+		handler = &privateServiceHandler{}
```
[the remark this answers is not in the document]
smcclure20: Elaborating a bit more from the above: I think this just reflects the underlying cloud abstractions. We are creating a private service connect, but to do that, we need the URL of a service attachment.

pkg/gcp/resources.go
```diff
@@ -110,62 +139,74 @@ func parseResourceUrl(resourceUrl string) (*resourceInfo, error) {
...
 	createClusterRequest := &containerpb.CreateClusterRequest{}
+	serviceAttachment := &ServiceAttachmentDescription{}
 	err := json.Unmarshal(resourceDesc, insertInstanceRequest)
 	if err == nil && insertInstanceRequest.InstanceResource != nil {
-		return &gcpInstance{}, nil
+		return &instanceHandler{}, nil
 	} else if err := json.Unmarshal(resourceDesc, createClusterRequest); err == nil && createClusterRequest.Cluster != nil {
-		return &gcpGKE{}, nil
+		return &clusterHandler{}, nil
+	} else if err := json.Unmarshal(resourceDesc, serviceAttachment); err == nil {
```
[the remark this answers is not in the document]
smcclure20: I think we should do type-specific checks like this in `getResourceInfo`. I'll add that.

pkg/gcp/resources.go
smcclure20: I thought it was nice to separate the initiation of the clients into its own function (saves a bunch of error checking and forces client errors to happen early), so if we keep that, we will have to enumerate which clients it uses anyways. I am not opposed to embedding the full thing, but I think these are fairly stable once created.

pkg/gcp/plugin.go
```diff
@@ -493,13 +496,33 @@ func (s *GCPPluginServer) _GetUsedAddressSpaces(ctx context.Context, req *paragl
 				resp.AddressSpaceMappings[i].AddressSpaces = append(resp.AddressSpaceMappings[i].AddressSpaces, *secondaryRange.IpCidrRange)
 			}
 		}
+
+		// Get addresses not associated with the vpc (might be used for PSCs)
+		listAddressesReq := &computepb.ListAddressesRequest{
+			Project: project,
+			Filter:  proto.String(fmt.Sprintf("labels.paraglider_ns eq %s", deployment.Namespace)), // TODO NOW: make the label name a constant
+		}
+		listAddressesResp := addressesClient.List(ctx, listAddressesReq)
```
seankimkdy, line 505: Is this query a superset of the for loop above where we query for the exact VPC/subnets for each deployment? If so, could we get rid of the above for loop? If we do both then I assume it would actually create duplicates?

seankimkdy (changes requested):

pkg/gcp/resources.go
```diff
@@ -53,35 +57,74 @@ type resourceNetworkInfo struct {
 	SubnetUrl   string
 	ResourceID  string
 	NetworkName string
+	Address     string
+}
+
+type ServiceAttachmentDescription struct {
```
praveingk, line 63: Since the definition of `ServiceAttachmentDescription` is defined in paraglider scope and not GCP, may be we need to define it somewhere, when we document private endpoint support. Just noting it here :)

praveingk (approved):

pkg/gcp/plugin.go
```diff
@@ -493,13 +496,33 @@ func (s *GCPPluginServer) _GetUsedAddressSpaces(ctx context.Context, req *paragl
 				resp.AddressSpaceMappings[i].AddressSpaces = append(resp.AddressSpaceMappings[i].AddressSpaces, *secondaryRange.IpCidrRange)
 			}
 		}
+
+		// Get addresses not associated with the vpc (might be used for PSCs)
+		listAddressesReq := &computepb.ListAddressesRequest{
+			Project: project,
+			Filter:  proto.String(fmt.Sprintf("labels.paraglider_ns eq %s", deployment.Namespace)), // TODO NOW: make the label name a constant
```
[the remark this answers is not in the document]
smcclure20: Thanks for catching this :)

pkg/gcp/plugin.go
smcclure20: Nope, these are separate resource types. In this one, I am searching for Addresses (IP adresses) vs. searching for VPCs and getting their address spaces. VPC address spaces are not Addresses (weirdly :) )

pkg/gcp/resources.go
smcclure20: Yeah, we definitely should include it in the docs about this.

seankimkdy (approved):

PR #419, opened 2024-07-06
Add AttachResource to Orchestrator

J-467:
- Using the CreateResource POST URL to handle AttachResource requests, so
- Created a handler to deal handle both Create and attach requests
- Added tests for attach resource in orchestrator
- Contributes to #170


pkg/paragliderpb/paraglider.proto
```diff
@@ -69,6 +69,10 @@ message ResourceDescriptionString {
     string description = 2;
 }
 
+message ResourceString {
```
smcclure20, line 72: Ah, okay now that I see this here, I think it is a bit out of place. Maybe define this in `orchestrator.go` and just call it `AttachResourceRequest` or something like that? It might be worthwhile doing the same for `ResourceDescriptionString` in the future too...

smcclure20 (changes requested):
Looks good! Just one nit!

pkg/paragliderpb/paraglider.proto
J-467: Sounds good. However, I don't think we should `AttachResourceRequest` for it. I think something more generic like `Resource`/`ResourceID / `ResourceWithID` for easy reusability might be more ideal? Thoughts? And we also have an `AttachResourceRequest` in the protobufs already in use

smcclure20: How about we go with `ResourceID`. Sorry for asking for further changes!

J-467: No worries at all. Will fix

J-467: Updated

smcclure20 (approved):

PR #420, opened 2024-07-08
Add attach resource to client

J-467:
- Adds the Attach resource to the client
- Includes test for attach resource
 - Minor refactor to use the protobuf objects when returning from fakes, instead of string maps
 - Child PR of #419


pkg/fake/orchestrator/rest/fake_orchestrator_rest_server.go
```diff
@@ -170,28 +170,53 @@ func (s *FakeOrchestratorRESTServer) SetupFakeOrchestratorRESTServer() string {
...
+				}
+			} else {
+				// Attach Resource (POST)
+				resource := &paragliderpb.ResourceString{}
+				err := json.Unmarshal(body, resource)
+				if err != nil {
+					http.Error(w, fmt.Sprintf("error unmarshalling request body: %s", err), http.StatusBadRequest)
+				}
+
+				err = s.writeResponse(w, &paragliderpb.AttachResourceResponse{Name: "validResourceName", Uri: resource.Id, Ip: "testIp"})
```
smcclure20, line 216: Nit: make these constants instead of having them inline?

smcclure20 (changes requested):

pkg/fake/orchestrator/rest/fake_orchestrator_rest_server.go
J-467: Done

smcclure20 (approved):

smcclure20 (approved):

PR #421, opened 2024-07-08
Refactor: modularize tests setup

J-467:
Modularizing test setup

### DIFF
--- a/pkg/client/client_test.go
+++ b/pkg/client/client_test.go
@@ -27,10 +27,14 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestGetPermitList(t *testing.T) {
+func setupClientWithFakeOrchestratorServer() Client {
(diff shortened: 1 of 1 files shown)


smcclure20 (approved):
Looks like it needs updated with other branches, but otherwise looks good!

pkg/paragliderpb/paraglider.proto
```diff
@@ -69,6 +69,10 @@ message ResourceDescriptionString {
     string description = 2;
 }
 
+message ResourceString {
```
smcclure20, line 72: I don't think this should be here, right?

smcclure20 (changes requested):

pkg/paragliderpb/paraglider.proto
J-467: Fixed

smcclure20 (approved):

PR #431, opened 2024-07-10
Update `GetUsedAddressSpaces` to use Paraglider vnet tags

### LINKED ISSUE #404: Update GetUsedAddressSpaces logic in Azure Plugin
Currently, the function checks the vnets with the paraglider prefix to determine if the vnet is "controlled" by Paraglider. Since we're introducing external vnets which can have different names, they may/will not follow this naming convention but may still be controlled by Paraglider

J-467:
- Currently, `GetUsedAddressSpaces` checks the vnet name prefix to determine if the vnet is managed by Paraglider. 
- Since we're supporting external vnet attachments, this PR updates the function to use the namespace tags attached to every Paraglider managed vnet, instead of just the name
- Fixes #404

### DIFF
--- a/pkg/azure/plugin.go
+++ b/pkg/azure/plugin.go
@@ -376,7 +376,7 @@ func (s *azurePluginServer) GetUsedAddressSpaces(ctx context.Context, req *parag
 			return nil, err
 		}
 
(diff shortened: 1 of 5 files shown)


smcclure20 (approved):
Thanks for fixing this!

Issue #187, opened 2024-04-05
[Infra] Run tests within devcontainer in GitHub Actions workflow


seankimkdy:
Our workflow file `workflows/build.yaml` specify a lot of versions of libraries like Go, `gotestsum`, Python, `protoc`, and other pre-installed tools. This causes us to keep track of versions in both our workflow file and devcontainer file. We should try to run our workflow within the devcontainer if possible. This may be of help https://github.com/devcontainers/ci/blob/main/docs/github-action.md.

divega:
Triage: @seankimkdy, we moved this to backlog, but if you think we should do it sooner, please assign it to yourself and move it to vnext.

seankimkdy:
https://github.com/devcontainers/ci could be the answer to this problem.

PR #433, opened 2024-07-10
[Infra] Ignore more paths to reduce extraneous builds

### LINKED ISSUE #259: Don't run integration tests on docs-only PRs
Per discussion with @seankimkdy, this is already true but filters on /docs. It would be nice to filter on file type so that changes to things like the CONTRIBUTING file do not trigger builds.

seankimkdy:
I updated our workflows to not run "Build and Test" and "Functional Test Approvals" for PRs that only change some files. The current reasoning behind each (listed under `paths-ignore` in the diff) is as follows.

- `.devcontainer/`: Not used in any of the build process, though this may change with #187
- `.githooks/`: same as above
- `docs/`: same as above (and already existed before this PR)
- `**.md`: excludes files such as README, CONTRIBUTING, CODE-OF-CONDUCT
- `**.rst`: not sure if this isn't covered anywhere that `docs/` doesn't cover since we don't use RST outside of docs but just as a safe measure

Resolves #259. @smcclure20 let me know your thoughts.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,14 +7,22 @@ on:
     tags:
       - v*
     paths-ignore:
+      - '.devcontainer/**'
+      - '.githooks/**'
       - 'docs/**'
+      - '**.md'
+      - '**.rst'
   pull_request:
     branches:
       - main
       - features/*
       - release/*
     paths-ignore:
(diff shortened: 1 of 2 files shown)


smcclure20 (approved):
LGTM! Thanks!

PR #437, opened 2024-07-15
[Azure] Integration test for Attach resource

### LINKED ISSUE #405: [Azure] Add integration test for AttachResource
- [x] Add integration test for attach resource function in plugin

_Originally posted by @J-467 in https://github.com/paraglider-project/paraglider/issues/170#issuecomment-2161210207_

J-467:
- Integration test for Attach Resource
- Closes #405


pkg/azure/plugin_integration_test.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func createVM(ctx context.Context, server *azurePluginServer, subscriptionId string, resourceGroupName string, namespace string, location string, name string) (*paragliderpb.CreateResourceResponse, error) {
+func createParagliderVM(ctx context.Context, server *azurePluginServer, subscriptionId string, resourceGroupName string, namespace string, location string, name string) (*paragliderpb.CreateResourceResponse, error) {
```
seankimkdy, line 35: Any reason this was renamed? If this clashes with another name, then a better name might be `createTestVM`. We don't have a concept known as a "Paraglider VM", so the method might be misleading.

Also, I prefer to keep names short as Go is already verbose.

pkg/azure/plugin_integration_test.go
```diff
@@ -360,3 +360,168 @@ func TestMultipleRegionsIntraNamespace(t *testing.T) {
...
+		SubscriptionID:    subscriptionId,
+		ResourceGroupName: resourceGroupName,
+		ResourceName:      externalVmName,
+	}
+	externalAddressSpace := "20.9.0.0/24"
+	azureHandler, err := azureServer.setupAzureHandler(resourceIdInfo, namespace)
+	require.NoError(t, err)
+
+	// Create Non-Paraglider Vnet
+	externalVnetParams := CreateVirtualNetworkParameters(ctx, vmLocation, externalVnetName, externalAddressSpace)
```
seankimkdy, line 418: This probably should have been addressed in a previous PR, but could we rename this to `GetVirtualNetworkParameters`? Create makes it sound like it's going to create the VNET.

And also, we shouldn't export this method unless there is a major reason to use this in other packages.

seankimkdy (changes requested):

pkg/azure/utils.go
```diff
@@ -529,3 +529,28 @@ func DoesVnetOverlapWithParaglider(ctx context.Context, handler *AzureSDKHandler
 
 	return false, nil
 }
+
+// CreateVirtualNetworkParameters creates and returns an instance of armnetwork.VirtualNetwork
+// with the specified parameters.
+//
+// Subnet address space is the same as the Vnet address space.
+func CreateVirtualNetworkParameters(ctx context.Context, location string, vnetName string, addressSpace string) armnetwork.VirtualNetwork {
```
seankimkdy, line 537: Could we rename this to GetVirtualNetworkParameters? Create makes it sound like it's going to create the VNET.

And also, we shouldn't export this method unless there is a major reason to use this in other packages.

seankimkdy (changes requested):

pkg/azure/plugin_integration_test.go
J-467: I was initially thinking of creating a function that returns a VM not created within Paraglider's vnet. However, I decided against that, so I would rename to as before

pkg/azure/utils.go
J-467: Alright

pkg/azure/plugin_integration_test.go
```diff
@@ -410,12 +412,28 @@ func TestAttachResourceIntegration(t *testing.T) {
...
+
+	// Find unused address spaces for external address
+	conn, err := grpc.NewClient(azureServer.orchestratorServerAddr, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		utils.Log.Printf("Azure Integration test: Could not dial the orchestrator")
+		return
+	}
+	defer conn.Close()
+	client := paragliderpb.NewControllerClient(conn)
+	reqAddressSpaces := make([]int32, 2)
```
seankimkdy, line 424: Why length of 2? I thought we only need one external address space, in which we can actually pass `nil`.

J-467: You're right. We do need only 1. I had 2 for debugging purposes. Fixed this now. Thanks Sean

smcclure20 (approved):
LGTM! Thanks!

seankimkdy (approved):

PR #438, opened 2024-07-16
[Doc] Attach resource doc

J-467:
Documentation for Attach Resource

### DIFF
--- a/docs/source/overview/api.rst
+++ b/docs/source/overview/api.rst
@@ -219,7 +219,8 @@ Note that a tag is automatically created for the resource with the name ``<names
 
                 * ``id``: Id of the resource as specified by the cloud provider
     
-Note: Create resource and Attach resource share the same API endpoint. The difference between the two POST requests is the request body. If the request body contains a ``description``, the request is considered as a create resource request. Otherwise, it is considered as an attach resource request. 
+..note:
+    Create resource and Attach resource share the same API endpoint. The difference between the two POST requests is the request body. If the request body contains a ``description``, the request is considered as a create resource request. Otherwise, it is considered as an attach resource request. 
 
 
 Permit List Operations
--- a/docs/source/overview/api.rst
+++ b/docs/source/overview/api.rst
@@ -178,8 +178,9 @@ Note that a tag is automatically created for the resource with the name ``<names
 Attach
 ^^^^^^
 
-Attaches a resource to Paraglider according to the resource ID and the specified cloud. 
-Note that a tag is automatically created for the resource with the name ``<namespace>.<cloud>.<name>`` after attachment (where ``name`` is the resource name provided).
+Attaches an exisiting resource to Paraglider according to the resource ID and the specified cloud. The resource should exist within a deployment associated with a namespace in Paraglider.
+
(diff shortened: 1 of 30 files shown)


docs/source/overview/api.rst
```diff
@@ -175,6 +175,53 @@ Note that a tag is automatically created for the resource with the name ``<names
...
+
+                    {
+                        "id": "resource-id"
+                    }
+
+                Parameters:
+
+                * ``id``: Id of the resource as specified by the cloud provider
+    
+Note: Create resource and Attach resource share the same API endpoint. The difference between the two POST requests is the request body. If the request body contains a ``description``, the request is considered as a create resource request. Otherwise, it is considered as an attach resource request.
```
seankimkdy, line 222: Let's make this into a callout

```suggestion
..note:
    Create resource and Attach resource share the same API endpoint. The difference between the two POST requests is the request body. If the request body contains a ``description``, the request is considered as a create resource request. Otherwise, it is considered as an attach resource request. 
```

docs/source/overview/api.rst
```diff
@@ -175,6 +175,53 @@ Note that a tag is automatically created for the resource with the name ``<names
                 * ``resource_name`` : name of the resource to be created in the Paraglider controller (note: this name will be scoped on cloud and namespace when stored)
                 * ``description``: JSON string describing the resource to be created (excluding networking details)
 
+Attach
+^^^^^^
+
+Attaches a resource to Paraglider according to the resource ID and the specified cloud.
```
seankimkdy, line 181: Should we make it clearer that the resource should already exist? And if there's any criteria for attaching resources?

seankimkdy (changes requested):

docs/source/overview/api.rst
J-467: Added that it should be within a Paraglider deployment. However, didn't want to add that the address spaces shouldn't overlap and permit rules restricitons as that might be too detailed. Thoughts?

seankimkdy: I think it's ok. I agree it might be a bit verbose, but we should definitely have a requirements list for what you can attach. Otherwise, there's no way for users to know what is attachable.

docs/source/overview/api.rst
```diff
@@ -175,6 +175,55 @@ Note that a tag is automatically created for the resource with the name ``<names
                 * ``resource_name`` : name of the resource to be created in the Paraglider controller (note: this name will be scoped on cloud and namespace when stored)
                 * ``description``: JSON string describing the resource to be created (excluding networking details)
 
+Attach
```
seankimkdy, line 178: Could we also add a warning callout that states this is for Azure only at the moment and other clouds are WIP?

As a tangent, do the other clouds return gracefully at the moment when using this?

seankimkdy (changes requested):

docs/source/overview/api.rst
J-467: Added the note that it's only supported by Azure currently. The attach API uses the create resource API route, so it would return gracefully. However, the error message may not always be "not supported". I'll create an issue for this.

praveingk (approved):

seankimkdy (approved):
LGTM sorry I missed this

PR #439, opened 2024-07-16
Pre-creation/attachment check at orchestrator level

### LINKED ISSUE #429: Check that resource tag name exists before attempting to create
Check that resource tag name exists before attempting to create the resource on the cloud

When using the CLI with glide create

J-467:
- Checks that a resource tag does not already exist before creating or attaching a resource
- Closes #429

### DIFF
--- a/pkg/orchestrator/orchestrator.go
+++ b/pkg/orchestrator/orchestrator.go
@@ -1127,6 +1127,10 @@ func (s *ControllerServer) handleCreateOrAttachResource(c *gin.Context) {
 		return
 	}
 
+	if err := s.preCreateOrAttach(c, resourceInfo); err != nil {
+		return
+	}
+
 	var resourceToCreate paragliderpb.ResourceDescriptionString
 	var resourceToAttach ResourceID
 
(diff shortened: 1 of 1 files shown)


pkg/orchestrator/orchestrator.go
```diff
@@ -1150,6 +1154,32 @@ func (s *ControllerServer) handleCreateOrAttachResource(c *gin.Context) {
 	}
 }
 
+// Check the resource is valid to be created or attached at the orchestrator level
+func (s *ControllerServer) preCreateOrAttach(c *gin.Context, resourceInfo *ResourceInfo) error {
```
seankimkdy, line 1158: I'm contemplating whether we need this before or within `resourceAttach`. We already have a `createTag` call within `resourceAttach`. Could the tag checking happen before that call instead of before the entire `resourceAttach`?

If we think one step further, should the `createTag` method actually check if there already exists a tag with a given

pkg/orchestrator/orchestrator.go
```diff
@@ -1127,6 +1127,10 @@ func (s *ControllerServer) handleCreateOrAttachResource(c *gin.Context) {
 		return
 	}
 
+	if err := s.preCreateOrAttach(c, resourceInfo); err != nil {
+		return
```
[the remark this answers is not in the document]
J-467: I'm not returning an error response because at any point an error occures in `preCreateOrAttach`, AbortWithStatusJSON is called to fill in the error message

pkg/orchestrator/orchestrator.go
J-467: As it is currently, the check tag checking via `preCreateOrAttach` happens before either creating or or attaching. I think it's a good idea to add the check to `createTag`. However, in this instance, we'd want that check to happen before we call the plugin's create or attach resource, to skip that process entirely if that tag exits. Hence, having the check only in `createTag` wouldn't fix the problem. Or we could maintain a duplicated call?

smcclure20: So the crux of the issue is here that we want two things: (1) that if the resource creation is going to fail, we don't create a tag for the resource and (2) if the tag creation is going to fail (because the tag already exists), we shouldn't find that out only *after* creating the resource. So to accomplish both of these, we need to check the tag is going to work before trying to create the resource (or vice versa, but that's much harder). So I think these have to be separate calls.

seankimkdy (approved):

PR #440, opened 2024-07-16
Enable IBM cloud tests

praveingk:


seankimkdy (changes requested):
@praveingk I think you need to get the secrets as environment variables here: https://github.com/paraglider-project/paraglider/blob/4eb8cf1234a70a885a38deb28a204e4e6003710b/.github/workflows/functional-tests.yaml#L22-L25

seankimkdy:
https://github.com/paraglider-project/paraglider/actions/runs/10087837840 is the manually triggered workflow run.

praveingk:
Seems like it failed due to "Failed to start redis server" , any ideas? It seems to work in my local env.

seankimkdy:
Updated run here: https://github.com/paraglider-project/paraglider/actions/runs/10151694203

pkg/ibm/vpn.go
```diff
@@ -434,6 +434,7 @@ func (c *CloudClient) DeleteVPN(VPNGatewayID string) error {
 			utils.Log.Printf("Failed to delete routes of VPN connection %v, during VPN deletion process, with error: %+v", *connection.ID, err)
 			return err
 		}
+		time.Sleep(10 * time.Second)
```
Cohen-J-Omer, line 437: @praveingk Consider replacing sleep with `pollVPNStatus(ID, False)`.
It returns true if the VPN is ready/deleted within the alloted time frame.

pkg/ibm/vpc.go
```diff
@@ -111,6 +112,15 @@ func (c *CloudClient) TerminateVPC(vpcID string) error {
...
 	}
+	// wait until vpc is deleted
+	for {
+		_, err = c.GetVPCByID(vpcID)
+		if err != nil {
+			break
+		}
+		time.Sleep(1 * time.Second)
+		fmt.Printf("Waiting for VPC to be deleted..\n")
+	}
```
Cohen-J-Omer, line 123: @praveingk can you please add a `maxAttempts`/`timeLimit` parameter to avoid an endless loop (resources deletion processes could hang for a long time on IBM cloud). Upon reaching the threshold value the method should return an error.
For example see `attachTag()` in `tagging.go`.

pkg/ibm/vpn.go
praveingk: Thanks Omer. I will try that, I am currently hitting some race conditions during deletion. I think polling status would help.

praveingk:
@Cohen-J-Omer At this point I am still facing errors due to VPN connection termination, due to routes still existing. I am still debugging this, on why its happening in GitHub actions, and not in my local env.

Issue #434, opened 2024-07-10
Add NATs/Public IPs for Internet Connections


smcclure20:
Currently, while adding a public IP to a permit list will result in the underlying network security rules allowing the traffic, if the instance does not have a public IP or the VPC does not have a NAT, the instance will be unable to communicate with the external endpoint. 

Accordingly, when a rule is added that allows a connection to a public IP, a NAT in the region should be provisioned if it is not already. This applies to all plugins.

seankimkdy:
@Cohen-J-Omer Do you have the capacity to take on this task? I completed this for Azure (#441) and GCP (#442) and they were fairly straightforward.

Cohen-J-Omer:
Hey @seankimkdy,
I'm familiar with the process (implemented it before using VpcV1 python SDK), but I'm currently needed elsewhere. If it's not urgent I believe I could implement it in the near future.

seankimkdy:
@Cohen-J-Omer Sounds good. Not urgent at the moment so take your time.

PR #441, opened 2024-07-22
[Azure] Add NAT gateways for outbound public IP addresses

seankimkdy:
Takes on Azure portion of #434. This allows resources to have outbound permit list rules with public IP addresses. When a permit list rule with a public IP address target is passed in to `AddPermitListRules`, a NAT gateway is spun up in the vnet where the resource resides. Similar to VPN gateways, the NAT gateway is created on demand and not before it is necessary.

> [!NOTE]  
> Initiating inbound connections is still not supported and can only be done if a user attaches a public IP address to the VM. Support for this may come later but hasn't been discussed.

I also refactored Azure's connectivity check methods such that it's more user friendly to those from other clouds.

GCP PR is coming soon after #417 is merged due to a lot of code overlap.


pkg/azure/plugin.go
```diff
@@ -173,10 +173,49 @@ func (s *azurePluginServer) AddPermitListRules(ctx context.Context, req *paragli
 
 		for i, peeringCloudInfo := range peeringCloudInfos {
 			if peeringCloudInfo == nil {
-				continue
-			}
-			address := rule.Targets[i]
-			if peeringCloudInfo.Cloud != utils.AZURE {
+				// Setup NAT gateway for public IP addresses
+				natGatewayLocation := *resourceVnet.Location
```
smcclure20, line 177: Nit: Maybe this should be its own function?

pkg/azure/test_utils.go
```diff
@@ -754,3 +779,123 @@ func getFakeVnetInLocation(location *string, addressSpace string) *armnetwork.Vi
...
+		if *resp.ConnectivityInformation.ConnectionStatus == armnetwork.ConnectionStatus("Reachable") {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
+
+func RunICMPConnectivityCheck(ctx context.Context, namespace string, subscriptionId string, resourceGroupName string, sourceVmName string, destinationIPAddress string, tries int) (bool, error) {
+	// Note that external ICMP
```
smcclure20, line 889: Unfinished comment?

smcclure20 (changes requested):
Just a couple nits! Thanks!

pkg/azure/test_utils.go
seankimkdy: Good catch!

smcclure20 (approved):

PR #442, opened 2024-07-23
[GCP] Add NAT gateways for outbound public IP addresses

seankimkdy:
Takes on GCP portion of #434. The NAT gateway requires a cloud router like the VPN gateway we use for multicloud connections. I've chosen to share the router between these two functionalities as GCP says there's no issues as long as BGP is not shared.

> You can reuse the same Cloud Router with different network connectivity products. However, each BGP session is unique to the network connectivity product (VLAN attachment, Cloud VPN tunnel, or router appliance instance) that you configure to use with Cloud Router. Different network connectivity products cannot use the same BGP session. 

(see https://cloud.google.com/network-connectivity/docs/router/how-to/create-router-vpc-network#establish_bgp_sessions)

Other minor things include a refactor of GCP connectivity test methods.


smcclure20 (changes requested):
Just a minor suggestion

smcclure20 (approved):

PR #443, opened 2024-07-24
[Infra] Add manual trigger for functional tests

seankimkdy:
This adds a manual trigger so that we can trigger the functional tests manually with the workflow file of a branch for the PR instead of pulling from main. No security concerns as only people with write access to the repo are allowed to manually trigger.

The caveat to this is that the checks shown at the bottom of the PR will still show failures from the auto triggered version. For any PRs that require changes to workflow files (which shouldn't be too many), we'll have to override the checks requirement and just merge as long as the manual triggered run passes tests.

### DIFF
--- a/.github/workflows/functional-tests.yaml
+++ b/.github/workflows/functional-tests.yaml
@@ -1,6 +1,12 @@
 name: Functional Tests
 
 on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


smcclure20 (approved):
Thank you for figuring this out!

PR #449, opened 2024-07-25
[AWS] Add GitHub Actions authentication

seankimkdy:
Doing this in a separate PR before the first AWS feature comes so that we don't have to manually trigger every time we want to run a workflow on that first feature PR.


seankimkdy:
Refer to https://github.com/paraglider-project/paraglider/actions/runs/10088320445 for manual trigger

smcclure20 (approved):

PR #454, opened 2024-07-30
[AWS] Resource creation for VM

### LINKED ISSUE #453: [AWS] Resource creation (VMs only)

seankimkdy:
Resolves #453. Some unique things/limitations about AWS

**Faking AWS Server for Unit Testing**
Unlike GCP and Azure, faking an HTTP server did not work. I used [middlewares](https://aws.github.io/aws-sdk-go-v2/docs/middleware/) instead which is a bit of a different concept. `fakeDeserializeMiddleware` should look a lot like `getFakeServerHandler` in GCP or Azure.

**Using Namespaces for Resource Cleanup during Integration Test**
AWS does not have resource groups in the sense that Azure does, nor does it have a nice isolation mechanism like GCP project does. There's no clean way to delete resources by a group, so I had to write a custom deletion by hand in `TeardownAwsTesting`. This is also why namespace has to be unique as there can be many different tests going on in the same account for AWS.

**VPC Quota Limit**
There's currently a limit of default 5 VPCs per region in AWS. This will not do as we tend to have more >5 ongoing PRs. We should ask for an [service quota increase](https://docs.aws.amazon.com/general/latest/gr/aws_service_limits.html).


pkg/aws/integration_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	req := &paragliderpb.CreateResourceRequest{
+		Deployment: &paragliderpb.ParagliderDeployment{
+			Id:        accountId,
+			Namespace: namespace,
+		},
+		Name:        instanceName,
+		Description: testInstanceJson,
+	}
+	resp, err := awsServer.CreateResource(context.TODO(), req)
```
smcclure20, line 56: Why the `TODO`?

pkg/aws/naming.go
```diff
@@ -0,0 +1,86 @@
...
+// getDescribeFilter returns a filter for a resource to use for getting (e.g., Describe...).
+func getDescribeFilter(namespace string, name string) []types.Filter {
+	return []types.Filter{
+		{Name: aws.String("tag:Name"), Values: []string{name}},
+		{Name: aws.String("tag:Namespace"), Values: []string{namespace}},
+	}
+}
+
+// getCreateTagSpecifications returns tag specifications for a resource to use for creating (e.g., Create...).
+func getCreateTagSpecifications(namespace string, name string, resourceType types.ResourceType) []types.TagSpecification {
```
smcclure20, line 40: This naming is just a bit awkward to me. Maybe something like `getIntitialParagliderResourceTags` or something like that?

pkg/aws/plugin.go
```diff
@@ -0,0 +1,258 @@
...
+		return nil, fmt.Errorf("resource description should not contain subnet ID")
+	}
+	if runInstancesInput.PrivateIpAddress != nil || runInstancesInput.EnablePrimaryIpv6 != nil ||
+		runInstancesInput.Ipv6AddressCount != nil || len(runInstancesInput.Ipv6Addresses) > 0 {
+		return nil, fmt.Errorf("resource description should not contain IP address information")
+	}
+
+	// Get region
+	availabilityZone := *runInstancesInput.Placement.AvailabilityZone
+	region := availabilityZone[:len(availabilityZone)-1]
```
smcclure20, line 68: Maybe this should wait until there's more implemented, but might this be helpful to have in a helper function?

pkg/aws/plugin.go
```diff
@@ -0,0 +1,258 @@
...
+			err = securityGroupExistsWaiter.Wait(ctx, &ec2.DescribeSecurityGroupsInput{
+				Filters: []types.Filter{
+					{Name: aws.String("vpc-id"), Values: []string{*createVpcOutput.Vpc.VpcId}},
+				},
+			}, 10*time.Second)
+			if err != nil {
+				return nil, fmt.Errorf("unable to wait for default security group: %w", err)
+			}
+
+			// Remove default inbound and outbound rules from default security group
```
smcclure20, line 127: Would it be easier to just put a deny-all rule at a higher priority than these defaults? That's what we do in Azure.

pkg/aws/plugin.go
```diff
@@ -0,0 +1,258 @@
...
+		subnetName := getSubnetName(req.Deployment.Namespace, region)
+		describeSubnetsOutput, err := ec2Client.DescribeSubnets(ctx, &ec2.DescribeSubnetsInput{
+			Filters: getDescribeFilter(req.Deployment.Namespace, subnetName),
+		})
+		if err != nil {
+			return nil, fmt.Errorf("unable to get subnets: %w", err)
+		}
+		if len(describeSubnetsOutput.Subnets) == 1 {
+			subnet = &describeSubnetsOutput.Subnets[0]
+			if subnet.AvailabilityZone != nil && *subnet.AvailabilityZone != availabilityZone {
```
smcclure20, line 181: Does this mean we need a subnet for each AZ? If so, shouldn't we add that to the filter for getting them or something? / Create a new subnet in each new AZ we create a resource in?

pkg/aws/plugin_test.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"testing"
+
+	orchestrator "github.com/paraglider-project/paraglider/pkg/fake/orchestrator/rpc"
+	"github.com/paraglider-project/paraglider/pkg/paragliderpb"
+	"github.com/paraglider-project/paraglider/pkg/utils"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCreateResource(t *testing.T) {
```
smcclure20, line 30: Should there be tests for the availability zone specific logic?

pkg/aws/test_utils.go
```diff
@@ -0,0 +1,301 @@
...
+var fakeInitializeMiddleware = middleware.InitializeMiddlewareFunc("FakeInput", func(
+	ctx context.Context, in middleware.InitializeInput, next middleware.InitializeHandler,
+) (
+	out middleware.InitializeOutput, metadata middleware.Metadata, err error,
+) {
+	// Pass request parameters as part of context for deserialize middleware
+	return next.HandleInitialize(context.WithValue(ctx, &requestContextKey{}, in.Parameters), in)
+})
+
+// fakeDeserizliaeMiddle is a middleware in the deserialize step for faking AWS API calls during tests.
```
smcclure20, line 87: Typo

smcclure20 (changes requested):
Thank you for figuring this all out! Looks great with a few notes!

pkg/aws/integration_test.go
seankimkdy: > TODO returns a non-nil, empty [Context](https://pkg.go.dev/context#Context). Code should use context.TODO when it's unclear which Context to use or it is not yet available (because the surrounding function has not yet been extended to accept a Context parameter).

https://pkg.go.dev/context#TODO

While I was writing unit tests for AWS, I actually ended up using these contexts to pass some information like the fake server state. It's unclear whether we'll have more of a use case for contexts, but I figured it was good to use this instead of `context.Background()` because the description of `context.TODO()` fits our current situation right now. We should probably have `context.TODO()` in the rest of the codebase, but it's low priority.

pkg/aws/plugin.go
seankimkdy: I don't think AWS has deny rules for security groups. From https://docs.aws.amazon.com/vpc/latest/userguide/security-group-rules.html#security-group-rule-characteristics

> You can specify allow rules, but not deny rules.

There are network ACLs (which have both allow and deny) but I believe they apply over an entire subnet? Let me know if I'm wrong about that though cus I'm not 100% sure.

pkg/aws/plugin.go
seankimkdy: Yea this was a little more involved than what I thought was needed for the initial version. I would need to split IP address into the different availability zones on the very first VM creation. This IP address range split cannot be changed after the first VM creation. Since the user needs to specify which availability zone they want to create a VM in, there may be an issue where a user wants to put a bulk of their VMs in a particular availability zone but it doesn't have any addresses left because the address range has been evenly split across all the other availability zones.

An argument could be made that user doesn't specify the particular availability zone but rather the region which is broader. I would also like this approach because then we can do the load balancing across different availability zones. The only question is where would the user specify the region? There is no region parameter in RunInstances ([source](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html)), only availability zone.

pkg/aws/plugin_test.go
seankimkdy: Good point. I added a test.

pkg/aws/plugin.go
smcclure20: Okay yeah, given this I think the option that makes the most sense (for now) is to create subnets as needed when VMs specify a new AZ and get a new address space for each. Since the address space allocation now allows you to get variable size prefixes, we can make them whatever size makes sense.

smcclure20 (approved):
Approved with the note that we should fix how AZs are handled when creating a VPC in a region. Per discussion with @seankimkdy, we should provision a subnet in each AZ on VPC creation by splitting the VPC address space into N equal chunks.

pkg/aws/plugin.go
seankimkdy: Issue at #474

PR #455, opened 2024-07-31
Update index.rst with some suggestions.

divega:
I just started reading this today again before sharing it with some people at Microsoft and I found a couple of things that I thought could be improved. Again :smile:


seankimkdy (approved):

praveingk (approved):
Looks good to me.

PR #493, opened 2024-11-03
Permit List check for visibility

J-467:
- Checking that the permit lists are configured properly and the targets are valid


pkg/azure/plugin.go
```diff
@@ -847,6 +847,211 @@ func (s *azurePluginServer) AttachResource(ctx context.Context, attachResourceRe
...
+			// If fixing, both inbound and outbound rules need to be deleted
+			if !shouldFix && visitedIps[peerIp] {
+				continue // Skip if the IP has already been visited and checked
+			}
+
+			if peeringCloudInfo == nil {
+				requireNAT = true
+			} else if peeringCloudInfo.Cloud == utils.AZURE {
+				// Ips don't can't be repeated, so if a deleted Ip is encountered again
+				// in a diff. rule, it's for conn. in the opposite direction and should be deleted
```
smcclure20, line 981: Not sure I understand this. Why would that imply its in the opposite direction?

pkg/azure/plugin.go
```diff
@@ -847,6 +847,211 @@ func (s *azurePluginServer) AttachResource(ctx context.Context, attachResourceRe
...
+				if shouldFix && deletedIps[peerIp] {
```
smcclure20, line 982: Shouldn't this check for membership in the map instead of assuming its there?

pkg/azure/plugin.go
```diff
@@ -847,6 +847,211 @@ func (s *azurePluginServer) AttachResource(ctx context.Context, attachResourceRe
...
+					err = handler.DeleteSecurityRule(ctx, *networkInfo.NSG.Name, rule.Name)
+					if err != nil {
+						utils.Log.Printf("An error occured while deleting security rule:%+v", err)
+						return nil, err
+					}
+					continue
+				}
+
+				// Get the URI for the peered resource
+				getIpReq := &paragliderpb.GetUriFromIpRequest{Ip: peerIp, Cloud: utils.AZURE}
```
smcclure20, line 992: The tags of all resources referenced in a rule are stored in the description. It might be better to query based on the tag instead of the IP?

smcclure20: In fact, I don't think there is a way to query by IP in the underlying database structure.

pkg/orchestrator/orchestrator.go
```diff
@@ -1528,6 +1701,31 @@ func (s *ControllerServer) DeleteValue(c context.Context, req *paragliderpb.Dele
 	return &paragliderpb.DeleteValueResponse{}, nil
 }
 
+func (s *ControllerServer) GetUriFromIp(c context.Context, req *paragliderpb.GetUriFromIpRequest) (*paragliderpb.GetUriFromIpResponse, error) {
```
smcclure20, line 1704: This is very inefficient, especially for large numbers of tags. If we want the ability to query like this, we need to change the underlying structure of the tag service. As mentioned above, I think you should query by tag instead.

smcclure20 (changes requested):
Looks generally good. Just a concern about how you use the tag service.

pkg/azure/plugin.go
J-467: It skipped me that we're storing the tags for the permit lists. I'll change this to use the tags instead

J-467: Also realizing for attached resources, our permit list targets may not have tags stored for them. In that case I'm guessing we can skip them?

pkg/azure/plugin.go
J-467: Inbound target : [10.1.1.1]
Outbound target: [10.1.1.1]

Since we don't support overlapping addresses, any 10.1.1.1 references the same resource (regardless of cloud). If the target's inbound rule has been deleted, the target's outbound rule should be deleted as well

pkg/azure/plugin.go
J-467: Do you mean, `if -, ok := deletedIps[peerIp]; ok {...}` ?

If yes, I think this code does the same because go returns the default value (i.e false) if the key is not in the map

J-467:
Updated version of this PR is at #503

PR #494, opened 2024-11-03
Allow rules for GetPermitList

J-467:
- `GetPermitList` checks if a rule is an allow rule before converting to Permit List
- However, it did this by checking if the prefix of the name is "paraglider...".
- For attached resources, the name of the rule may not conform to paraglider naming requirement

### DIFF
--- a/pkg/azure/plugin.go
+++ b/pkg/azure/plugin.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v4"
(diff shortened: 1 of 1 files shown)


J-467:
@smcclure20 If you could check this out, that'll be great

pkg/azure/plugin.go
```diff
@@ -90,7 +89,8 @@ func (s *azurePluginServer) GetPermitList(ctx context.Context, req *paragliderpb
 
 	// get the NSG rules
 	for _, rule := range nsg.Properties.SecurityRules {
-		if !strings.HasPrefix(*rule.Name, denyAllNsgRulePrefix) && strings.HasPrefix(*rule.Name, paragliderPrefix) {
+		// Ignore default Azure rules with priority above maxPriority
+		if *rule.Properties.Access == allowRule && *rule.Properties.Priority <= maxPriority {
```
smcclure20, line 93: There shouldn't be any deny rules, right? So is the first part of this check even necessary?

J-467: There's the deny-all rule for inbound and outbound rules

J-467: Removing the <= maxPriority because Rules, doesn't contain the default Azure rules, so no need for that

pkg/azure/plugin.go
```diff
@@ -90,7 +89,8 @@ func (s *azurePluginServer) GetPermitList(ctx context.Context, req *paragliderpb
 
 	// get the NSG rules
 	for _, rule := range nsg.Properties.SecurityRules {
-		if !strings.HasPrefix(*rule.Name, denyAllNsgRulePrefix) && strings.HasPrefix(*rule.Name, paragliderPrefix) {
+		// Ignore default Azure rules with priority above maxPriority
```
divega, line 92: Change below looks good to me, but I don’t understand how this code comment applies or where it is implemented.

divega (commented):
LGTM except for one comment.

pkg/azure/plugin.go
J-467: This is for the attach functionality, where we're getting the permit lists of attached resources

PR #500, opened 2024-11-15
[GCP] Fix for private service connect bug

### LINKED ISSUE #495: [GCP] Failed to create private service resource
Creating the file `gcp_psc_ep.json` with the below contents:

```
{
   "url": "projects/my-project-name/regions/us-east4/serviceAttachments/httpd-lb-service",
   "region": "us-east4"
}
```

and invoking:

```
glide resource create gcp vpe gcp_psc_ep.json
```

produced the below message:

```
failed to create resource: Put "http://localhost:8080/namespaces/default/clouds/gcp/resources/vpe": EOF
```

which seems to cause the orchestrator to crash:

```
panic: runtime error: index out of range [0] with length 0

goroutine 68 [running]:
github.com/paraglider-project/paraglider/pkg/gcp.(*privateServiceHandler).readAndProvisionResource(0xc000c749c0, {0x30b0308, 0xc000882c60}, 0xc000e0e5a0, {0xc000eae300, 0x1c}, 0xc0003965b0, {0x4b0b240, 0x0, 0x0})
        /home/weit/paraglider/pkg/gcp/resources.go:631 +0x1b9
github.com/paraglider-project/paraglider/pkg/gcp.ReadAndProvisionResource({0x30b0308, 0xc000882c60}, 0xc000e0e5a0, {0xc000eae300, 0x1c}, 0xc0003965b0, {0x4b0b240, 0x0, 0x0}, 0xc000e0e600)
        /home/weit/paraglider/pkg/gcp/resources.go:243 +0x11a
github.com/paraglider-project/paraglider/pkg/gcp.(*GCPPluginServer)._CreateResource(0xc000dabee0, {0x30b0308, 0xc000882c60}, 0xc000e0e5a0, 0xc000e0e600)
        /home/weit/paraglider/pkg/gcp/plugin.go:438 +0x1665

```

I updated the below gcp resources.go method to return 1

```
// Get the subnet requirements for a private service connect attachment
func (r *privateServiceHandler) getNumberAddressSpacesRequired() int {
	return 1
}
```

re-built and run glide command again - which then produced the below error:

```
failed to create resource: Request failed with status code 400: {"error":"rpc error: code = Unknown desc = unable to read and provision resource: unable to insert address: googleapi: Error 400: Invalid value for field 'resource.purpose': 'PRIVATE_SERVICE_CONNECT'. The type/purpose values are invalid for subnetwork address."}
```

after updating addressPurpose constant with the below value: 

```
addressPurpose                = "GCE_ENDPOINT"
```

I was able to create a gcp endpoint with `glide resource create` CLI:

```
~/paraglider$ glide resource create gcp vpe gcp_psc_ep.json
Creating resource: vpe
Resource Created.
tag: default.gcp.vpe
uri: https://www.googleapis.com/compute/v1/projects/my-project-name/regions/us-east4/forwardingRules/para-vpe
ip: 10.0.0.3
```

I would like to ask whether I am missing something or perhaps a fix should be applied.

Thanks.

smcclure20:
Resolves #495 

Root issues:
* Some of the original code was just simply wrong (changing request fields after sending the request)
* Google service connections require Global IPs and Global Forwarding rules

Future issues:
* There is a lot of repeated code here - we can probably clean it up
* This and other resources should take advantage of the fact that the orchestrator can now give variable-sized address spaces


praveingk (approved):
Looks good!

praveingk (approved):

pkg/gcp/clients.go
```diff
@@ -202,9 +226,15 @@ func (c *GCPClients) Close() {
 	if c.addressesClient != nil {
 		c.addressesClient.Close()
 	}
+	if c.globalAddressesClient != nil {
+		c.globalAddressesClient.Close()
```
divega, line 230: For my education, are these clients functional after calling Close? E.g., do they provide methods to re-open or do they automatically re-establish anything needed for operation if you make any subsequent call? 

Otherwise, should we be setting fields to nil explicitly, to allow for a functional client to be populated again via GetOrCreate methods? 

Perhaps it's ok if we only use Close when we are shutting down the serice or going to recycle the entire struct?

pkg/gcp/resources.go
```diff
@@ -649,15 +653,15 @@ func (r *privateServiceHandler) getResourceInfo(ctx context.Context, resource *p
...
+		region = globalRegion
 	}
 
 	return &resourceInfo{Region: region, NumAdditionalAddressSpaces: r.getNumberAddressSpacesRequired(), ResourceType: privateServiceConnectTypeName, Name: resource.Name}, nil
 }
 
 // Get the subnet requirements for a private service connect attachment
 func (r *privateServiceHandler) getNumberAddressSpacesRequired() int {
-	return 0
+	return 1 // Only used for GCP services (TODO: Change this to depend on that)
```
divega, line 664: Should we create an issue to track this in more detail, or are we ok with tracking with TODOs?

divega (commented):
Just a couple of newbie questions.

pkg/gcp/clients.go
smcclure20: I don't think they can be reopened after closing. We do (at this point in time) call `Close()` at the end of every request and recreate the struct on every request. Good point that maybe we should reuse as we can and only call close on the shutdown of the service. I can add an issue for this.

pkg/gcp/resources.go
smcclure20: Will add an issue.

PR #502, opened 2024-11-17
[Observability] Permit Lists config check

J-467:
- Checks for Permit List Configuration and Network existence
- Network existence checks for the VNET and it's provisioning state
- Permit List Configuration checks for no interleaved allow and deny rules. And that NSG exists and successfully provisioned

check-permit-list-conf


smcclure20 (approved):
Looks good to me! I might have some questions about the CheckCodes being test-specific, but I can address that in the base PR

PR #511, opened 2025-02-07
Optional arguments for `make test`

smcclure20:
Adds the ability to specify individual packages when running `make test`.

### DIFF
--- a/build/test.mk
+++ b/build/test.mk
@@ -17,8 +17,7 @@ GOTEST_OPTS ?=
 GOTEST_TOOL ?= gotestsum $(GOTESTSUM_OPTS) --
 endif
 
-# Optionally test just one package
-# By default, does not test internal directory packages if a specific package is specified
+# Optionally test an individual package
 INTERNAL_PKG ?= ...
 PKG ?= ...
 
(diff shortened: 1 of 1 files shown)


build/test.mk
```diff
@@ -17,8 +17,30 @@ GOTEST_OPTS ?=
 GOTEST_TOOL ?= gotestsum $(GOTESTSUM_OPTS) --
 endif
 
+# Optionally test an individual package
```
divega, line 20: Nit: wondering if something like this would work and look simpler? Otherwise, I would still prefer indentation to make the code easier to follow:

``` make

# Set package directories with explicit conditionals.
ifeq ($(PKG),...)
  GOTEST_PKG :=
else
  GOTEST_PKG = ./pkg/$(PKG)
endif

ifeq ($(INTERNAL_PKG),...)
  GOTEST_INTERNAL_PKGS :=
else
  GOTEST_INTERNAL_PKGS = ./internal/$(INTERNAL_PKG)
endif

# If neither directory was set, default to testing all.
ifeq ($(strip $(GOTEST_PKG) $(GOTEST_INTERNAL_PKGS)),)
  GOTEST_PKG = ./pkg/...
  GOTEST_INTERNAL_PKGS = ./internal/...
endif
```

smcclure20: Thank you, this is much better than what I had.

divega (approved):

Issue #510, opened 2025-02-04
[GCP] Only provide additional address space for Google services for PSCs


smcclure20:
As noted in [this PR comment](https://github.com/paraglider-project/paraglider/pull/500#discussion_r1924560284), currently the `getNumberAddressSpacesRequired()` method for GCP private endpoints returns 1. This is necessary for Google services which need to have a separate address than anything currently existing. However, for non-Google services this extra address space will not be used. This function should be updated to consider this and return 0 when it is a non-Google service.

PR #513, opened 2025-02-20
Doesn't Allocate Address Space for Non-Google Services

kellyyyeh:
Returns 0 for non-Google services in `getNumberAddressSpacesRequired()`.


smcclure20 (approved):
Looks great! Thanks for the fix!

smcclure20:
Resolves #510

PR #514, opened 2025-02-27
Return error if tag doesn't exist

Ankita39Infy:
Signed-off-by: Ankita<Ankita.sharma39@infosys.com>

### DIFF
--- a/internal/cli/glide/tag/get/get.go
+++ b/internal/cli/glide/tag/get/get.go
@@ -67,6 +67,9 @@ func (e *executor) Execute(cmd *cobra.Command, args []string) error {
 	if e.resolveFlag {
 		tagMappings, err := c.ResolveTag(args[0])
 		if err != nil {
+			if isTagNotFound(err){
+				return fmt.Error("tag '%s' dose not exist", args[0])
+			}
 			return err
 		}
 
@@ -83,4 +86,11 @@ func (e *executor) Execute(cmd *cobra.Command, args []string) error {
 	}
 
 	return nil
+}
+func isTagNotFound(err error) bool {
+	var httpErr *http.ResponseError
+	if error.As(err, &httpErr) && httpErr.Response.StatusCode !=200{
+		return true
+	}
+	return false
 }


internal/cli/glide/tag/get/get.go
```diff
@@ -84,3 +87,10 @@ func (e *executor) Execute(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+func isTagNotFound(err error) bool {
+	var httpErr *http.ResponseError
+	if error.As(err, &httpErr) && httpErr.Response.StatusCode !=200{
```
smcclure20, line 92: There are many errors that may return from the controller for this request, not just that the tag was not found. In fact, I am not sure that a non-existent tag creates an error. Instead, it just returns the name and everything else is empty. That said, perhaps we should change the orchestrator to throw a specific error when the tag does not exist. @divega any thoughts?

smcclure20 (changes requested):
Thank you so much for the PR!

You found a behavior in the code that I think we should consider carefully how to handle (perhaps the logic should reside server-side, not client-side). 

While we discuss that, I will also not that the signature on the commits has to be when you commit locally (ie, when you run `git commit -m "your message"` that you include `-s` to make it ``git commit -m "your message" -s`. We also have a githook you can install on your machine to make this happen locally (explained [here](https://paragliderproject.io/developers/contributing.html#developer-certificate-of-origin:~:text=Developer%20Certificate%20of-,Origin,-%C2%B6)).

internal/cli/glide/tag/get/get.go
divega: Catching up here... @smcclure20, you mentioned in chat (and I was able to see following the code) that under the hood we are calling HGETALL command against Redis, which doesn't return an error but an empty map if the key does not exist. I will assume that this same behavior is surfaced by the c.ResolveTag() call here. 

To move forward with the PR, I suggest:
1. In general, writing a test that tries to retrieve an non-existent tag to observe how things behave should be the first step. The test should be included in the PR and will help figure out what the right solution is. 
2. Assuming the behavior is what I see in the code (we probably return an empty map), we need to decide if this is the correct behavior, not just for the CLI but for the REST API. When an operation tries to retrieve a single element that doesn't exist, it seems better to return an error message conveying "not found" instead of returning nothing. While the PR has good intentions, I suggest we consider implementing a change at the API (server) level rather than at the CLI.
3. Once the API returns the correct status (I'll assume HTTP 404 would be correct, even if ResolveTag is a POST call and not GET), we can do something on the CLI to output a clear message. 

Does this make sense?

divega: If the contract of the corresponding tag service API is also to find a single item by key, I would suggest when not found, it should also return something that conveys "not found". If on the other hand the tag service API is designed to return n results, returning 0 results should be enough.

Not sure if this is helpful. I am not in front of my computer right now, and I don't even remember if the tag service API is rest or gRPC.

divega: Confirmed just that with Sarah that the tag service API is gRPC based, which means instead of 404 we need to come up with our own way to signal "not found".

divega: Hi @Ankita39Infy, we appreciate your efforts on this PR. Are you planning to continue working on it? If you're busy, we understand. If you would like to chat about the changes we are suggesting, please let us know and we will try to find a way.

PR #517, opened 2025-04-05
[GCP] [Draft] Compliance Tests of a Resource with Paraglider

kellyyyeh:
- Adds functions and tests to determine the compliance of existing GCP resources with paraglider for attachment
- Criteria for compliance:
  - [x] Validate that resource is valid and is a supported type
  - [x] Ensure VPC address isn't overlapping within the deployment
  - [x] Ensure that permit list of the VNet has a deny all rule with the lowest priority (i.e. highest priority number)
  - [x] Ensures that all rules with higher priority than the deny all rule are allow rules
  - [x] Any deny rule must be a deny all rule. All rule verifications disregard the default rules created by GCP with priority > 65534

### DIFF
--- a/pkg/gcp/firewalls.go
+++ b/pkg/gcp/firewalls.go
@@ -261,23 +261,12 @@ func CheckFirewallRulesCompliance(firewalls []*computepb.Firewall) (bool, error)
 	return true, nil
 }
 
-// Iterates and locates the deny-all firewall rule with the lowest priority
-func findDenyAllRule(firewalls []*computepb.Firewall) *computepb.Firewall {
-	var denyAllRule *computepb.Firewall
-	lowestSeenPriority := int32(lowestPriority)
-
-	for _, fw := range firewalls {
-		if isDenyAllRule(fw) && fw.GetPriority() >= lowestSeenPriority {
-			lowestSeenPriority = fw.GetPriority()
-			denyAllRule = fw
-		}
-	}
-
-	return denyAllRule
-}
-
 // Checks if a firewall rule is a deny-all rule
 func isDenyAllRule(fw *computepb.Firewall) bool {
+	if fw == nil {
+		return false
+	}
+
 	return len(fw.Denied) > 0 &&
 		len(fw.Allowed) == 0 &&
 		strings.ToLower(fw.Denied[0].GetIPProtocol()) == "all"
--- a/pkg/gcp/firewalls.go
+++ b/pkg/gcp/firewalls.go
@@ -31,6 +31,8 @@ const (
 	firewallRuleDescriptionPrefix = "paraglider rule" // GCP firewall rule prefix for description
 	targetTypeTag                 = "TAG"
 	targetTypeAddress             = "ADDRESS"
+	highestPriority               = 0
+	lowestPriority                = 65534
 )
 
 // Maps between of GCP and Paraglider traffic direction terminologies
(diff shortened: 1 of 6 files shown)


pkg/gcp/firewalls.go
```diff
@@ -229,3 +231,59 @@ func parseDescriptionTags(description string) []string {
...
+
+	return true, nil
+}
+
+// Iterates and locates the deny-all firewall rule with the lowest priority
+func findDenyAllRule(firewalls []*computepb.Firewall) *computepb.Firewall {
+	var denyAllRule *computepb.Firewall
+	lowestSeenPriority := int32(lowestPriority)
+
+	for _, fw := range firewalls {
```
smcclure20, line 264: Nit: If other functions don't need this helper, maybe it should be integrated into the above. That way, we only iterate through the list of firewalls once (finding the lowest priority rule and making sure it matches Deny-All) and that all others are allows).

pkg/gcp/resources.go
```diff
@@ -243,13 +269,199 @@ func GetFirewallTarget(ctx context.Context, resourceInfo *resourceInfo, netInfo
...
+	isCompliant, err := CheckFirewallRulesCompliance(firewallRules)
+	if err != nil || !isCompliant {
+		return nil, nil, fmt.Errorf("Firewall rules are not compliant: %w", err)
+	}
+	
+	return resourceInfo, networkInfo, nil
+}
+
+// GetVPCAddressSpace retrieves the address space for a given VPC network
+func GetVPCAddressSpace(ctx context.Context, project, vpcName string, clients *GCPClients) ([]string, error) {
```
smcclure20, line 323: This seems generally useful. Maybe we should create an issue for finding other uses of this function.

pkg/gcp/resources.go
```diff
@@ -243,13 +269,199 @@ func GetFirewallTarget(ctx context.Context, resourceInfo *resourceInfo, netInfo
...
+	// 	Network: vpcName,
+	// 	Project: resourceInfo.Project,
+	// }
+
+	// getNetworkResp, err := networksClient.Get(ctx, getNetworkReq)
+	// if err != nil {
+	// 	return false, fmt.Errorf("failed to get paraglider vpc network: %w", err)
+	// }
+
+	req := &paragliderpb.GetUsedAddressSpacesRequest{
```
smcclure20, line 406: I am a bit confused with this code. I would expect this function to call `GetUsedAddressSpaces` on the *orchestrator* not the cloud plugins (see paraglider.proto) which just takes an empty message. That way, all address spaces from all clouds are returned (not just the ones in GCP). Then, we can iterate through those address spaces and compare with the address spaces in the virtual network (by getting it). I believe there is a common util function for seeing if two address spaces overlap. Happy to explain in person if that would help.

smcclure20 (commented):
Looks good so far! Only really one significant comment.

I also didn't review the commented code for now, but will do once its ready.

Thanks for all the hard work!

pkg/gcp/resources_test.go
```diff
@@ -198,6 +201,72 @@ func TestIsValidResource(t *testing.T) {
 	assert.Equal(t, resource.Name, resourceInfo.Name)
 }
 
+func TestValidateResourceCompliesWithParagliderRequirements(t *testing.T) {
```
smcclure20, line 204: Maybe add a negative case where it fails?

smcclure20 (changes requested):
Looks great! Just a couple ideas on how to make it a bit cleaner / have better coverage

pkg/gcp/firewalls.go
```diff
@@ -229,3 +231,52 @@ func parseDescriptionTags(description string) []string {
...
 	return getParagliderNamespacePrefix(namespace) + "-fw"
 }
+
+// 1. A deny-all rule exists at the lowest priority (highest number)
+// 2. All higher-priority rules (lower numbers) are allow rules
+func CheckFirewallRulesCompliance(firewalls []*computepb.Firewall) (bool, error) {
+	var denyAll *computepb.Firewall
+	var highestDenyPrio = int32(lowestPriority)
+	for _, firewall := range firewalls {
+		println("highest seen", highestDenyPrio)
```
smcclure20, line 241: Nit: I think you accidentally left this in

pkg/gcp/firewalls.go
```diff
@@ -229,3 +231,52 @@ func parseDescriptionTags(description string) []string {
...
+		if len(firewall.Denied) > 0 {
+			if isDenyRule(firewall) {
+				if strings.ToLower(firewall.Denied[0].GetIPProtocol()) == "all" {
+					denyAll = firewall
+				}
+				if firewall.GetPriority() < int32(highestDenyPrio) {
+					highestDenyPrio = firewall.GetPriority()
+					println("new highest seen", highestDenyPrio)
```
smcclure20, line 249: Same as above

pkg/gcp/firewalls.go
```diff
@@ -229,3 +231,52 @@ func parseDescriptionTags(description string) []string {
...
+	return true, nil
+}
+
+// Checks if a firewall rule is a deny rule
+func isDenyRule(firewall *computepb.Firewall) bool {
+	if firewall == nil {
+		return false
+	}
+
+	// Checks that DestinationRanges is set to "0.0.0.0/0"
```
smcclure20, line 274: This is a requirement to be a deny *all* rule, not just a deny rule. I think the easiest way to fix this is to put the following check into this function and then rename it to `isDenyAllRule`
```
if strings.ToLower(firewall.Denied[0].GetIPProtocol()) == "all" {
					denyAll = firewall
				}
```

smcclure20: Essentially, in order to be a deny rule, the only thing that has to be true is that len(firewall.Denied)>0, which you already check in the for loop above. So you can use that to see if it is a deny rule, then check if it is a deny-all rule, and otherwise keep your logic the same.

smcclure20 (changes requested):

pkg/gcp/firewalls_test.go
```diff
@@ -0,0 +1,158 @@
...
+				Denied:          []*computepb.Denied{{IPProtocol: proto.String("all")}},
+				DestinationRanges: []string{"192.168.0.0/16"},
+			},
+			expect: false,
+		},
+		{
+			name: "No rule",
+			expect: false,
+		},
+		{
```
smcclure20, line 146: Repeated test?

smcclure20 (approved):
Just one nit!

PR #518, opened 2025-04-07
Modify for version v0.1.0

praveingk:


smcclure20 (approved):

smcclure20:
Per discussion with @divega and others during the community meeting, can we add "preview" to the version name to indicate it isn't production-ready?

divega:
> Per discussion with @divega and others during the community meeting, can we add "preview" to the version name to indicate it isn't production-ready?

@smcclure20, I think we can wait until we are so close to 1.0.0 that we want to name our releases something like `1.0.0-preview.1`. While we are on 0.x releases we don't need to append pre-release identifiers.

praveingk:
@smcclure20 If you are ok, we can mention preview in the release notes, rather than mention it in the version name.

PR #520, opened 2025-04-07
[Docs] Explain and reference feature flags in docs

smcclure20:


docs/source/examples/controller-setup.rst
```diff
@@ -59,6 +59,42 @@ This file contains all information needed to spin up each of the microservices.
...
+* ``attachResourceEnabled``: Enables attaching existing resources (not created with Paraglider) to the deployment (ie, using the :ref:`attach operation <attach_resource>`).
+* ``kubernetesClustersEnabled``: Enables creating Kubernetes clusters in the deployment (ie, providing a cluster description to the :ref:`create resource operation <create_resource>`).
+* ``privateEndpointsEnabled``: Enables creating private endpoints (eg, Private Endpoints in Azure, Private Service Connect in GCP) in the deployment (ie, providing a cluster description to the :ref:`create resource operation <create_resource>`).
+
+Below is an example of how to set the feature flags in the configuration file:
+
+.. code-block:: yaml
+
+    featureFlags:
+        azure:
```
divega, line 81: I didn't realize it would look like this. I thought it was just a flat list. I thought the purpose of feature flags was only to protect users from undefined behavior when trying to use a feature with a plugin that didn't support it yet. 

I believe we need to be able to combine plugins with experimental support for a new feature with unmodified plugins that are completely unaware of the same feature. If it was a flat list, that would mean the unaware plugin would just ignore the feature flag and then the behavior when trying to use the feature will be undefined. 

If we keep this tree structure then we can make it part of the contract that plugins should throw when they see a feature flag set to true that they don't recognize. Is that what we have today?

docs/source/project/roadmap.rst
```diff
@@ -5,11 +5,11 @@ Roadmap
 
 This pages lists some of the features that are planned for the future of Paraglider.
 
-Features:
+Features
 ^^^^^^^^^^
 .. card::
 
-    Get the IBM plugin caught up with the other plugins in feature completeness. 
+    Get the IBM plugin caught up with the other plugins in feature completeness. [TODO: Maybe remove?]
```
divega, line 12: Is this TODO something you want to resolve before we merge the PR? What do we need to know to decide?

docs/source/examples/controller-setup.rst
smcclure20: I created an issue to track this: https://github.com/paraglider-project/paraglider/issues/523

divega (approved):

PR #535, opened 2025-05-01
[Orchestrator] Allow multiple rules to be added at once when applying to a tag

smcclure20:
I found this bug by attempting to add a rule with the `--ping` flag to a tag rather than a resource (eg, `default.azure.vm1` instead of `azure vm1`). This results in two rules being sent in the payload to the orchestrator. While this is fine when applied to the function for resources, the one for tags only parsed one rule. 

I updated it to accept a list of rules (as the other APIs do)


divega (commented):
Do we need to add tests for this still?

divega (approved):
Approved. I am ok with creating an issue to track adding tests.

divega (approved):

PR #541, opened 2025-06-06
[Dev] Fixing dev container creation in json

J-467:
- Updated dev container to use go version 1.23
- Updated dev container references from protoc and tmux-apt-get


smcclure20 (approved):
Thank you for catching this!

PR #542, opened 2025-06-08
[Lint] Linter version upgrade

J-467:
- Migrated the golangci linter from v1 to v2. Migration updated the `.golangci.yaml` to the new version
- Run `make lint` to fix formatting issues on files
- Most changes are from "if....else if" to switch cases
- Several other fixes involved changes error messages to start with lower cases and removing punctuations from the end of error message

### DIFF
--- a/pkg/ibm/security_group.go
+++ b/pkg/ibm/security_group.go
@@ -191,10 +191,8 @@ func (c *CloudClient) translateSecurityGroupRuleGroupRuleProtocolICMP(
 	if err != nil {
 		return nil, err
 	}
-	isEgress := false
-	if *ibmRuleIcmp.Direction == outboundType {
-		isEgress = true
-	}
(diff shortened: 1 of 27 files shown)


pkg/ibm/security_group.go
```diff
@@ -504,7 +500,7 @@ func IBMToParagliderRules(rules []SecurityGroupRule) ([]*paragliderpb.PermitList
 
 	for _, rule := range rules {
 		if rule.PortMin != rule.PortMax {
-			return nil, fmt.Errorf("SG rules with port ranges aren't currently supported")
+			return nil, fmt.Errorf("sG rules with port ranges aren't currently supported")
```
smcclure20, line 503: Nit: since this is an acronym, it looks odd now. Maybe we should write it out?

smcclure20 (approved):
Thanks for updating this! Just a very minor nit.

